PR #22, opened 2023-07-23
Circuitzoo

Krastanov:


Krastanov:
@adrianariton , @Barosandu , just for future reference for you, notice how I have started abstracting away the various sub-circuits. In particular, see the new CircuitZoo file and how the firstgenrepeater was modified to use it.

PR #23, opened 2023-08-02
readded double selection

Barosandu:


src/CircuitZoo/CircuitZoo.jl
```diff
@@ -37,6 +37,23 @@ struct Purify2to1 <: AbstractCircuit
     end
 end
 
+struct Purify3to1 <: AbstractCircuit
+    leaveout::Symbol
```
Krastanov, line 41: May we remove this for now, until the correct set of parameters is derived. For the moment it can be just an empty struct without any parameters

src/CircuitZoo/CircuitZoo.jl
```diff
@@ -37,6 +37,23 @@ struct Purify2to1 <: AbstractCircuit
...
+    function Purify3to1(leaveout)
```
Krastanov, line 42: and then there will not be a need for a special constructor either

Krastanov:
may we also keep the example files out for the moment and add them later in a separate pull request - - the examples are part of the documentation and I would like to avoid modifying them, rather just making a new one with the more sophisticated circuit

Krastanov:
If I understand correctly, #24 is replacing this one. Closing for the moment

PR #24, opened 2023-08-02
Qvis 3

Barosandu:
Removed oarametrization from 3to1 struct and from the purification func


src/CircuitZoo/CircuitZoo.jl
```diff
@@ -37,6 +37,13 @@ struct Purify2to1 <: AbstractCircuit
     end
 end
 
+struct Purify3to1 <: AbstractCircuit
+    leaveout::Symbol
+    function Purify3to1()
```
Krastanov, line 42: I am not quite sure this constructor really works. Something like `struct Purify3to1 <: AbstractCircuit end` would be simpler and should work fine.

src/CircuitZoo/CircuitZoo.jl
```diff
@@ -57,4 +64,37 @@ function (circuit::Purify2to1)(purifiedL,purifiedR,sacrificedL,sacrificedR)
...
+    apply!((sacrificedR1,purifiedR),gate)
+
+
+    apply!((sacrificedR1,sacrificedR2),gate)
+    apply!((sacrificedL1,sacrificedL2),gate)
+
+    measa1 = project_traceout!(sacrificedL1, basis1)
+    measb1 = project_traceout!(sacrificedR1, basis1)
+
+
```
Krastanov, line 81: There is quite a bit of extra whitespace here. Could you clean it up?

src/CircuitZoo/CircuitAbstraction.jl
Krastanov, line 1: could you remove this empty file from the pull request?

firstgenrepeater-05.purifier.mp4
Krastanov, line 1: Could you remove this from the pull request? It looks like a video file generated while you were testing things.

examples/firstgenrepeater/setup.jl
Krastanov, line 1: Could you remove this file? Same as the comment from the previous pull request: this is a modification of the example file, which is used as a tutorial. If we are doing this, we should have it as a separate example, not modifying the already working one.

adrianariton:
I am going to link my fork here, and we'll both keep working on that one, as it is cleaner. I have begun implementing the 3 3to1 circuits, and i will create a test file for them too later today. Basically a doubleselection circuit is a superposition of two singleselection (2to1) circuits.

adrianariton:
We will work here #26

PR #25, opened 2023-08-03
Remove isfree and add locks to each slot

Krastanov:


Krastanov:
@adrianariton @Barosandu , quite a few things related to the locks are now simplified. More simplification work needs to happen though

Krastanov:
also, be aware that the version number was bumped to 0.3, so you might need to do a resolve or update

PR #30, opened 2023-08-15
Added color to register net plot depending on the fidelity of entanglement + Added directional metadata

adrianariton:
1. This option only allows for plotting of the fidelity colorwise
2. Should i add another parameter specifying what observable i should plot to make it more general?


adrianariton:
Using it here: https://github.com/adrianariton/QuantumFristGenRepeater @Krastanov

Krastanov:
I like the new directed edge metadata, no need for changes there.

I have the impression that the two-qubit coloring can currently lead to crashes if we have bigger entangled states. Could you add a couple of tests showing what happens if you try to plot something like:

```
r = Register(3)
initialize!(r[1:3], (Z1⊗Z1⊗Z1+ Z2⊗Z2⊗Z2)/sqrt(2))
```

Could you rename the keyword argument to something like `twoqubitobservable=nothing` and make checks on `isnothing(twoqubitobservable)` so that the user can do something like `registerplot(..., twoqubitobservable=projector(state))`?

Krastanov:
I will also suggest separating the gray line plotting from the colored observable plotting. Let's have both, so that the logic is not too intermingled. Basically, copy whatever part of the gray line plotting you need and cordon it off with `!isnothing(twoqubitobservable)`. That would make refactoring the code down the line much easier. It would also make extending it to more complicated observables much easier.

Krastanov:
to avoid things breaking you probably will need a guard along the lines of `nsubsystems(state) == 2` inside of the `!isnothing(twoqubitobservable)` block

adrianariton:
I asdded the code as example, i will get to writing the readme and fixing the other issues you mentioned tommorow!

adrianariton:
Regarding the twoqubit observable,
Wouldnt it be actually cleaner to be able to plot all kinds of entasngled states (with more than 2 qubits)?

Im thinking of taking a function as parameter: `qubitprojector(n::Integer)`, which returns the projector for an n qubit state, and then simply projecting onto it. Is that ok? Or should we stick only with the 2 qubit approach?

adrianariton:
For some reason if i dont specify the color for the more thqan 2 qubit systems, and i put it equal to `NaN`, it does not work for the WGLmakie sim, but it does for the simple recording.

adrianariton:
@Krastanov i added the approach we discussed today with twoqubit links. 
Please review my code :)

ext/QuantumSavoryMakie/QuantumSavoryMakie.jl
```diff
@@ -57,13 +63,32 @@ function Makie.plot!(rn::RegisterNetPlot{<:Tuple{RegisterNet}}) # TODO plot the
...
+                        push!(regs[], regs[][end])
+                        push!(twoqubitstate_links[], twoqubitstate_links[][end])
+                    end
                 else
                     juststarted = false
                 end
             end
         end
+
+        println("-------$(length(twoqubitstate_links[]))--$(length(regs[]))")
```
Krastanov, line 82: looks like debugging info that was not deleted; please remove it

examples/fullstack/1_entangler_purifier_console.jl
```diff
@@ -0,0 +1,69 @@
...
+    # record the simulation progress
+    step_ts = range(0, time, step=0.1)
+    record(fig, "1_firstgenrepeater_$(length(registersizes))nodes.$(PURIFICATION ? "entpurif$(USE)to1" : "entonly").mp4", step_ts, framerate=10, visible=true) do t
+        run(sim, t)
+        notify(obs)
+        ax.title = "t=$(t)"
+    end
+end
+
+# TODO LIST
```
Krastanov, line 63: it seems there are a few places in the examples/fullstack folder that still have todos. Could you separate the example folder changes from the ext & src folder changes, so we can merge the ext & src work while the rest is being finished?

adrianariton: Yep, these are actually from a while ago and I did them (marked as Done). I will remove the lines shortly

PR #33, opened 2023-08-25
Adding fullstack example

adrianariton:


adrianariton:
@Krastanov line 13 in Sim.jl should be changed from
```
retina_scale = 1
```
to 
```
retina_scale = 2
```
if  ran on a Mac with retina display that supports double pixelling (if not changed, the figure resolution will be halved).
And left as such otherwise.

If left

Krastanov:
Thanks, this is very helpful!

I see you have a compat bound for WGLMakie. Probably related to https://github.com/MakieOrg/Makie.jl/issues/3161 ?

I will try to give a review later today

Krastanov:
There seems to be a session sharing issue (similar to the one originally reported in your CSSMakieLayout initial version). If I enable logs and open a second page `http://localhost:8888/1`, the logs from the simulation in the first page end up visible also on the second page. These should be separate sessions.

adrianariton:
Yep! Didn't notice that. I will fix is ASAP.

Krastanov:
No worries! It is not urgent and I will be sending more comments throughout the day anyway

adrianariton:
I think it should be working now

adrianariton:
I'll just leave my TODOs here so i remember them:
- add limits to waittime distrib. plot, and also plot it on menufig1
- cleanup first plot a little bit
- add log on both interfaces
- redo the landing page so it matches the updated version

adrianariton:
@Krastanov i added the README and Docs (you will notice that for sumtypes they are still comments, because i get an error saying they are not documentable).

Krastanov:
This was a very useful experiment, which informed our current approach, namely the use of `query` and tags by various protocols.

There are a few other useful developments here that we have not merged (the neat html makie layout, detailed networking diagrams, etc).

PR #34, opened 2023-08-26
Added Superdense endcoding and decoding circuits to CircuitZoo

Abhishek-1Bhatt:


Abhishek-1Bhatt:
Running the circuits from this PR is somehow returning inconsistent results:
```julia
using QuantumSavory
using QuantumSavory.CircuitZoo: SDEncode, SDDecode

##Set up an entangled bell pair
ra = Register(1)
rb = Register(1)

initialize!(ra[1], Z1)
initialize!(rb[1], Z1)

apply!(ra[1], H)
apply!((ra[1], rb[1]), CNOT)

message = [1, 1]

SDEncode()(ra, message)

rec = SDDecode()(ra, rb)
```
versus when simply running a program without using any registers:
```julia
using QuantumSavory
using QuantumClifford, QuantumOpticsBase

function alice_node1()
    message = rand(0:1, 2)
    println("Sending message: $message")
    state = S"XX
              ZZ"
    if message[1] == 1
        state = P"ZI"*state
    end
    if message[2] == 1
        state = P"XI"*state
    end
    return state
end

function bob_node2()
    state = alice_node1()
    apply!(state, tCNOT, [2,1])
    apply!(state, tHadamard,[2])
    msg = Ket(state)
    println("Message Received by Bob: $msg")
end

bob_node2()
```
Does there seem to be any mistake with the encoding or decoding circuit or could there be a bug somewhere?

Krastanov:
could you try to investigate what is happening step by step

the register has a staterefs property which you can use to see what state is being stored in it

```
julia> r = Register(3); initialize!(r[1], X1);

julia> r.staterefs[1].state[]
Ket(dim=2)
  basis: Spin(1/2)
 0.7071067811865475 + 0.0im
 0.7071067811865475 + 0.0im
```

You can also try the clifford representation given that you already have a related test

```
julia> r = Register(3, CliffordRepr()); initialize!(r[1], X1);

julia> r.staterefs[1].state[]
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z
𝒮𝓉𝒶𝒷
+ X
```

I will not always be able to review quickly after you submit a pull request. Please proceed with trying to investigate and fix even if I have not reviewed.

src/CircuitZoo/CircuitZoo.jl
```diff
@@ -743,4 +743,73 @@ function (circuit::PurifyExpedientNode)(purifiedL,sacrificedL)
...
+julia> SDDecode()(regA, regB)
+```
+"""
+
+struct SDDecode <: AbstractCircuit
+end
+
+function (circuit::SDDecode)(regA, regB)
+    apply!((regB[1], regA[1]), CNOT)
+    apply!(regB[1], H)
```
Abhishek-1Bhatt, line 809: Everything works fine upto this line, but when we apply H to Bob's qubit, the result is the same as what we would get if we applied H to Alice's qubit on paper, i.e., something like 
```julia
Ket(dim=4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
  0.49999999999999994 + 0.0im
 -0.49999999999999994 + 0.0im
 -0.49999999999999994 + 0.0im
  0.49999999999999994 + 0.0im
```
which gives us the random results.
But if we apply H to `regA[1]` , which is Alice's qubit here, it returns the correct results. I am not sure why it behaves this way, opposite to what we get when we solve it on paper, but it resolves the issue here

Krastanov: Could you elaborate a bit? What is the state before the wrong step, what is the step that seems to break things, and what is the state after that step (and what should it be)?

Abhishek-1Bhatt: For the final part of the decoder circuit, we want to apply H to the second qubit (Bob's qubit), and H is a 1 qubit operator. But under the hood our register stores a 4 element ket vector which (if the classical message is `1 1`) would be:
```julia
Ket(dim=4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
                 0.0 + 0.0im
                 0.0 + 0.0im
  0.7071067811865475 + 0.0im
 -0.7071067811865475 + 0.0im
```
To apply the symbolic H to the second qubit, we call the `apply!` method at https://github.com/QuantumSavory/QuantumSavory.jl/blob/09fe3aa93d0daa4d0964feb606a795a1274ad566/src/baseops/apply.jl#L12
Here(https://github.com/QuantumSavory/QuantumSavory.jl/blob/09fe3aa93d0daa4d0964feb606a795a1274ad566/src/baseops/apply.jl#L21) the `state_indices` parameter of our ket determines whether the final operator would be `I⊗H`(correct) or `H⊗I`(wrong) given state_indices is 1 or 2 respectively. Since, Bob's qubit is the 2nd qubit of the subsystem `state_indices` is set to 2 and it finally hits the `apply!` methods in QuantumOpticsBase.jl. So, because of `state_indices` being 2, the final operator [here](https://github.com/qojulia/QuantumOpticsBase.jl/blob/fe00cb9fd15d867ccd37bdcf8ef9826da3149893/src/apply.jl#L2) (in QuantumOpticsBase.jl) becomes `H⊗I` which gives us the wrong answer in my previous comment.

So, [sorry for the huge description : ) ] to be short, the `embed` method [here](https://github.com/qojulia/QuantumOpticsBase.jl/blob/fe00cb9fd15d867ccd37bdcf8ef9826da3149893/src/apply.jl#L2) expects to be provided with the index where we want to put the identity operator(index 1) instead of the index of the qubit on which we're performing the single qubit operation(index 2), as described by its docstring [here](https://github.com/qojulia/QuantumOpticsBase.jl/blob/fe00cb9fd15d867ccd37bdcf8ef9826da3149893/src/operators.jl#L22)
Hence, I think we have a bug in this line
https://github.com/QuantumSavory/QuantumSavory.jl/blob/09fe3aa93d0daa4d0964feb606a795a1274ad566/src/baseops/apply.jl#L21
and `state_indices` should be a vector of all the indices of the state excluding the index of our qubit

Abhishek-1Bhatt:
Needs the next QOBase release

PR #35, opened 2023-08-26
Some usability improvements for `project_traceout!()`

Abhishek-1Bhatt:
Allows `project_traceout()` to accept Tuples of possible basis states. Also added a more descriptive error message when the state to be measured is not normalized.

### DIFF
--- a/src/backends/quantumoptics/quantumoptics.jl
+++ b/src/backends/quantumoptics/quantumoptics.jl
@@ -37,7 +37,7 @@ function observable(state::Union{<:Ket,<:Operator}, indices, operation)
     expect(op, state)
 end
 
-function project_traceout!(state::Union{Ket,Operator},stateindex,psis::Union{Tuple{ZBasisState, ZBasisState},Vector{<:Ket}})
+function project_traceout!(state::Union{Ket,Operator},stateindex,psis::Union{Tuple,Vector{<:Ket}})
     if nsubsystems(state) == 1 # TODO is there a way to do this in a single function, instead of _overlap vs _project_and_drop
         _overlaps = [_overlap(psi,state) for psi in psis]
         branch_probs = cumsum(_overlaps)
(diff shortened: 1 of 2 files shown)


src/backends/quantumoptics/quantumoptics.jl
```diff
@@ -37,11 +37,13 @@ function observable(state::Union{<:Ket,<:Operator}, indices, operation)
     expect(op, state)
 end
 
-function project_traceout!(state::Union{Ket,Operator},stateindex,psis::Vector{<:Ket})
+function project_traceout!(state::Union{Ket,Operator},stateindex,psis::Union{Tuple,Vector{<:Ket}})
```
Krastanov, line 40: let's use `Base.AbstractVecOrTuple` here (which is almost the same union), to keep things consistent

Krastanov:
@Abhishek-1Bhatt , could you investigate what I broke here?

src/backends/quantumoptics/quantumoptics.jl
```diff
@@ -37,11 +37,13 @@ function observable(state::Union{<:Ket,<:Operator}, indices, operation)
     expect(op, state)
 end
 
-function project_traceout!(state::Union{Ket,Operator},stateindex,psis::Vector{<:Ket})
+function project_traceout!(state::Union{Ket,Operator},stateindex,psis::Base.AbstractVecOrTuple{Ket})
```
Abhishek-1Bhatt, line 40: So, the problem is that the symbolic basis states like `Z1`, `X2` etc aren't direct subtypes of `Ket`, but `SpecialKet` [here](https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/b95474d90664259800a132d30241c4560f27d028/src/QSymbolicsBase/predefined.jl#L5). 
Should I just replace it with 
```julia
function project_traceout!(state::Union{Ket,Operator},stateindex,psis::Base.AbstractVecOrTuple{<:Union{Ket, QuantumSymbolics.SpecialKet}})
```

PR #37, opened 2023-08-29
Improvements to the CircuitZoo purification circuits (fixes #32 and #31)

Barosandu:
All i have left is to check the varargs length
*edit* done

PR #45, opened 2023-09-23
Symbolic Objects for Local and Remote entanglement generation processes

Abhishek-1Bhatt:


Krastanov:
May we reorganize things a bit. Could you make two folders inside of the newly created StatesZoo folder, one for each of the two new objects. And then keep the matlab code for each in the corresponding folder. Let's also have a name that describes a bit more the "why" not the "what".

Krastanov:
something like
```
StatesZoo
- cascaded_source
-- precursor_code (that contains a readme explaining what was done to the code)
- entanglement_swap
-- precursor_code (again with explanation of what was done to it)
```

Keep in mind that these two things are very different types of processes and states, they should not be lumped together.

Also, keep the express statements local, do not put them in a completely different file.

Krastanov:
We need to figure out how to name these objects.

For the cascaded source, Saikat and Dirk refer to it as "ZALM source" for Zero-added-noise Multiplexed. It comes from https://arxiv.org/abs/2206.03670 and https://arxiv.org/abs/2206.03670 
So let's call the symbolic object `ZALMpair` and let's have it printed out as `ρᶻᵃˡᵐ`.

When you set up its doc string, please leave in `[](@cite)` references to these two papers. In terms of how to do the references, they need to be added as bibtex to the bibfile in the doc src folder. See for instance this pull request and the bibtex and docstrings in it https://github.com/QuantumSavory/QuantumSavory.jl/pull/37/files

The best way to get up-to-date bibtex references is from google scholar.

Krastanov:
For the midswap folder, let's have two separate symbolic objects defined there. Maybe for names we can use `DualRailMidSwapBell` and `SingleRailMidSwapBell`. Not quite sure about the names, please feel free to suggest better ones. Not quite sure for their unicode/latex representation either. Maybe ρᵈʳᵐˢ and ρˢʳᵐˢ but again, please feel free suggesting something better.

Krastanov:
In terms of how to do the symbolics, check out this file and how "SpecialKet" is done there https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/main/src/QSymbolicsBase/predefined.jl

The issue is that we do not really have the same for density matrices. I think a good way to do this is to make `abstract type AbstractTwoQubitState <: AbstractTwoQubitState end` (that probably should be in QuantumSymbolics) and then make a `struct ZALMPair <: AbstractTwoQubitState` (in QuantumSavory).

Abhishek-1Bhatt:
The `cascaded_source` folder has two functions for generating density matrices, one in the spin-spin basis which outputs a 4x4 density matrix and another in fock basis which outputs a 36x36 matrix, currently I have implemented the symbolic object ρᶻᵃˡᵐ being expressed as the one in spin-spin basis. Let me know what we should name the other one, it would be a small change. 

Also, in the above comment the part 
> I think a good way to do this is to make abstract type AbstractTwoQubitState <: AbstractTwoQubitState end

was not clear, so for now I did `abstract type AbstractTwoQubitState <: AbstractTwoQubitOp end` , let me know if you feel we should do it differently.

Then, I was not sure if we need the const `cascaded_source_basis` as a symbolic object, so its currently left as such, let me know what you think would be the best way to represent it.

src/StatesZoo/StatesZoo.jl
```diff
@@ -0,0 +1,56 @@
+module StatesZoo
+
+using QuantumSymbolics, QuantumOpticsBase
+using QuantumSymbolics: withmetadata, @withmetadata, Metadata
+import QuantumSymbolics: express_nolookup
+
+export cascaded_source_photonic, cascaded_source_spin, midswap_dual_rail, midswap_single_rail, ZALMpair
```
Krastanov, line 7: let's export only the symbolic objects (the structs), nothing else

src/StatesZoo/zalm_pair/cascaced_source_swapped_with_memories/readme.md
```diff
@@ -0,0 +1,139 @@
...
+fidelity=0.5*(M(2,2)+M(3,3)+M(2,3)+M(3,2));
+```
+
+## Code Dependencies
+- ret_cXY.m's are required for cascaded_source_spin.m
+
+
+## Reference
+1. [P. Dhara, S. J. Johnson, C. N. Gagatsos, P. G. Kwiat, S. Guha, Heralded Multiplexed High-Efficiency Cascaded Source of Dual-Rail Entangled Photon Pairs Using Spontaneous Parametric Down-Conversion. Phys. Rev. Appl. 17, 034071 (2022).](https://arxiv.org/abs/2107.14360)
+2. [K. C. Chen, et al., Zero-Added-Loss Entangled-Photon Multiplexing for Ground- and Space-Based Quantum Networks. Phys. Rev. Appl. 19, 054029 (2023).](https://arxiv.org/abs/2206.03670)
```
Krastanov, line 138: Reference 1 and 2 also matter. In `ZALMPair` we should say something along the lines "cascaded source (from 1 and 2) that is then stored in spin memories (from 3)". Could you update the references to include 1 and 2?

src/StatesZoo/zalm_pair/zalm_pair.jl
```diff
@@ -0,0 +1,588 @@
...
+    
+    return [m11 0 0 0 ; 0 m22 m23 0 ; 0 m23' m33 0 ; 0 0 0 m44]
+end
+
+
+"""
+Generate symbolic object for the spin-spin density matrix for a 
+cascaded source swapped with emissive spin memories from the paper [prajit2023entangling](@cite)
+"""
+@withmetadata struct ZALMpair <: AbstractTwoQubitState
```
Krastanov, line 568: To distinguish from the photonic one, let's call it ZALMSpinPair

src/StatesZoo/zalm_pair/zalm_pair.jl
```diff
@@ -0,0 +1,588 @@
...
+    Ns::Float64
```
Krastanov, line 569: Could you document what each of these parameters means? Prajit has that explained in his readmes -- basically just copy that, for all symbolic structs.

Krastanov:
could you add a few tests, something along the lines of:

```
reg = Register(2)
initialize!(reg[1:2], symbolicstate)
observable(reg[1:2], some_observable)
```

just to make sure that the API is not broken

src/StatesZoo/zalm_pair/cascaced_source_swapped_with_memories/readme.md
```diff
@@ -0,0 +1,139 @@
+# Cascaded Source Swapped With Emissive Memories
+
+## Overview
+This folder contains the functions required to generate the spin-spin density matrix for a cascaded source swapped with emissive memories. The functions are written in MATLAB. The functions are:
+- cascaded_source_photonic.m : Generates the photonic density matrix for a cascaded source from the papers Refs. 1 and 2.
+- cascaded_source_spin.m : Generates the spin-spin density matrix for a cascaded source swapped with emissive spin memories from the papers Ref. 3.
```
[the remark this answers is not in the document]
Abhishek-1Bhatt: Yes, the function for `photonic` density matrix has been translated to Julia, but there is no symbolic object for it right now. Same with the `basis`, I have declared it as a `const` in the module file `StatesZoo.jl` but its not used anywhere.

PR #49, opened 2023-10-12
Update make.jl

Abhishek-1Bhatt:


Krastanov:
I think the way to fix these missing references is to add a copy of the QuantumSymbolics API docs (basically an extra autodoc declaration for them in the API listing).

Krastanov:
it seems it still does not build correctly due to the same external references

Abhishek-1Bhatt:
the issue is with these docstrings in QuantumSymbolics: https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/7085f72484fd285b2c74a55673c9bfcc696312fa/src/QSymbolicsBase/QSymbolicsBase.jl#L64
and 
https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/7085f72484fd285b2c74a55673c9bfcc696312fa/src/QSymbolicsBase/express.jl#L9
They don't show up in the QuantumSymbolics docs either https://quantumsavory.github.io/QuantumSymbolics.jl/stable/API/

Krastanov:
What is the status of this?

Krastanov:
Converting this to a draft. Convert it back when it is ready.

PR #50, opened 2023-10-18
Separate normalized and unnormalized objects

Abhishek-1Bhatt:
Separated the symbolic objects into normalized and unnormalized, added a dispatch of `tr` from `LinearAlgebra.jl` for computing the trace of the unnormalized objects and fixed tests according to the new naming convention


src/StatesZoo/single_dual_rail_midswap/single_dual_rail_midswap.jl
```diff
@@ -94,17 +129,67 @@ julia> observable(r[1:2], Z⊗Z)
...
+end
+
+function express_nolookup(x::DualRailMidSwapBellN, ::QuantumOpticsRepr)
+    data = midswap_dual_rail(x.eA, x.eB, x.gA, x.gB, x.Pd, x.Vis)
+    return SparseOperator(_bspin⊗_bspin, Complex.(data/tr(data)))
+end
+
+## trace
+
+tr(x::SingleRailMidSwapBellU) = Float64(tr(express(x).data))
```
Krastanov, line 194: lets remove these -- unless it is some special constant, it does not make sense to implicitly turn a symbolic object into a float

src/StatesZoo/zalm_pair/zalm_pair.jl
```diff
@@ -609,9 +609,70 @@ juilia> observable(r[1:2], Z⊗Z)
...
\ No newline at end of file
+end
+
+function express_nolookup(x::ZALMSpinPairN, ::QuantumOpticsRepr)
+    data = cascaded_source_spin(x.Ns, x.gA, x.gB, x.eAm, x.eBm, x.eAs, x.eBs, x.eD, x.Pd, x.Pdo1, x.Pdo2, x.VisF)
+    return SparseOperator(_bspin⊗_bspin, Complex.(data/tr(data)))
+end
+
+## trace
+tr(x::ZALMSpinPairU) = Float64(tr(express(x).data))
```
Krastanov, line 678: same here

test/test_stateszoo_api.jl
```diff
@@ -1,15 +1,44 @@
...
+r_drmsU = Register(2)
+initialize!(r_drmsU[1:2], drmsU)
+@test ! iszero(observable(r_drmsU[1:2], Z⊗Z))
+
+r_drmsN = Register(2)
+initialize!(r_drmsN[1:2], drmsN)
+@test ! iszero(observable(r_drmsN[1:2], Z⊗Z))
+
+@test tr(zalmU) < 1.0
+@test tr(express(zalmN).data) ≈ 1
```
Krastanov, line 38: on the other hand, these should be exactly equal to 1 thanks to a special dispatch rule (so that `express` does not need to be used, nor `.data`)

Abhishek-1Bhatt: Wouldn't we still need to use express to convert it to `QuantumOptics` formalism so that [this](https://github.com/qojulia/QuantumOpticsBase.jl/blob/d2e4c8e98f3b7c0b38ead6434e9ffb0c5af1aaa4/src/operators_dense.jl#L186) specialised method can be called

Krastanov: I would prefer if we do not rely on quantumoptics.jl at all for that.

In symbolic CAS tools usually 1 and 0 are treated more specially than other numbers. For the normalized density matrices here it is guaranteed that the trace is 1, so I would like to have a special-cased method for them that completely shortcuts any expression functionality.

`@test  tr(rho) = tr(express(rho))` should work for them.

Please add these special-cased methods for `tr` and I should be able to merge.

Krastanov: @Abhishek-1Bhatt , bump

PR #53, opened 2023-10-31
Added QuantumChannel

Abhishek-1Bhatt:
Added QuantumChannel using `DelayQueue` from `ConcurrentSim.jl`.


src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+    elseif !isnothing(Δt) # if both are not nothing
+        uptotime!(rref.reg.staterefs[rref.idx], rref.reg.stateindices[rref.idx], qc.background, Δt)
+    end
+
+    put!(qc.queue, rref)
+end
+
+# should we mandate that rref_rec.reg is not initialized beforehand
+# should other register attributes of the sent regref be copied over like reprs[idx], backgrounds[idx], accesstimes[idx], env and locks[idx]
+@resumable function Base.take!(env, qc::QuantumChannel, rref_rec::RegRef)
```
Abhishek-1Bhatt, line 75: Should we check whether the register slot receiving the qubit is always uninitialized? Also, do we need to copy other information related to the qubit like the background noise associated with it, i.e., reprs(idx), backgrounds(idx), locks(idx) etc. ?

Krastanov: Yes, we should check that the receiving register is empty (same check as the one currently done in `initialize!`).

And no, all of these properties should not be copied. They are properties of the location (channel or register), not properties of the state stored in that location.

Krastanov:
Looking at the examples in ConcurrentSim, I do not think that `take!` and `put!` should be `@resumable`: https://github.com/JuliaDynamics/ConcurrentSim.jl/blob/8591bff85edb62556be142f141dafa506e0f7fae/src/resources/delayed_stores.jl#L56

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+    trait::Qubit
+    queue::ConcurrentSim.DelayQueue{RegRef}
+    background::Any
+end
+
+function QuantumChannel(queue::ConcurrentSim.DelayQueue{RegRef}, background=nothing)
+    QuantumChannel(Qubit(), queue, background)
+end
+
+function Base.put!(qc::QuantumChannel, rref::RegRef, Δt=nothing)
```
Krastanov, line 61: What is the meaning of `Δt` here? If it is how long something stays in the channel, then it should be a property of the channel. It should not be possible to vary for how long something stays in the channel (at least for now, that is a weird feature that we should avoid until everything else is fleshed out).

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+    if xor(isnothing(qc.background), isnothing(Δt))
+        throw(ArgumentError(lazy"""
+        Either both background and Δt should be nothing or both should be initialized to appropriate values
+        """))
+    elseif !isnothing(Δt) # if both are not nothing
+        uptotime!(rref.reg.staterefs[rref.idx], rref.reg.stateindices[rref.idx], qc.background, Δt)
+    end
+
+    put!(qc.queue, rref)
```
Krastanov, line 70: The `rref` should be "emptied" when the `put!` happens, not when the `take!` resolves, because the user might want to immediately reuse the `rref` for something else.

The cleanest way to do that right now is probably on each put! to create a temporary register:

```
uptotime!(rref, now(qc.queue.environment))
channel_reg = Register(1, [qc.background], [qc.trait])
swap!(rref, channel_reg[1])
uptotime!(channel_reg[1], qc.queue.delay)
put!(qc.queue, channel_reg)
```

then in the `take!` you should have

```
uptotime!(channel_reg[1], now(qc.queue.environment))
swap!(...)
```

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,34 @@
...
+regA = Register(2)
+regB = Register(2)
+initialize!((regA[1], regB[2]), bell)
+
+sim = Simulation()
+
+# Delay queue for quantum channel 
+queue = DelayQueue{RegRef}(sim, 10.0)
+
+qc = QuantumChannel(queue)
```
Krastanov, line 16: there should be a constructor that does not require you to provide an explicit `DelayQueue`

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,34 @@
...
+
+@process alice_node(sim, qc)
+@process bob_node(sim, qc)
+
+run(sim)
+
+sref = regB.staterefs[1]
+
+# the above code puts both the qubits of the state in the same register
+@test sref.registers[1] == sref.registers[2]
```
Krastanov, line 34: we should have a few tests that double-check that the time evolution due to backgrounds is properly evaluated, i.e. that all the necessary `uptotime!` are called internally.

src/quantumchannel.jl
Abhishek-1Bhatt: The `Δt` here is used by `uptotime!` to determine how long the background process acts on the transported state. Should it be channel attribute?

src/quantumchannel.jl
Abhishek-1Bhatt: Yeah, that makes sense!

test/test_quantumchannel.jl
Abhishek-1Bhatt: Do you mean other than the current constructor or should we modify the current constructor?

src/quantumchannel.jl
Krastanov: The latency of the channel, stored in the DelayQueue instance is already giving you that number.

test/test_quantumchannel.jl
Krastanov: a second constructor

src/QuantumSavory.jl
```diff
@@ -29,6 +30,9 @@ export Qubit, Qumode, QuantumStateTrait,
     CliffordRepr, QuantumOpticsRepr, QuantumMCRepr,
     UseAsState, UseAsObservable, UseAsOperation,
     AbstractBackground
+export QuantumChannel, take!, put!
```
Krastanov, line 33: do we need to export `take!` and `put!`? They are in `Base`

src/quantumchannel.jl
```diff
@@ -0,0 +1,86 @@
...
+  if !isnothing(qc.background)
+      uptotime!(channel_reg[1], qc.queue.delay)
+  end
+
+  put!(qc.queue, channel_reg)
+end
+
+
+@resumable function Base.take!(env, qc::QuantumChannel, rref_rec::RegRef)
+  if isassigned(rref_rec.reg, rref_rec.idx)
```
Krastanov, line 80: could you also add an `isassigned(::RegRef)` method, given that this here looks a bit silly (decomposing the reference before checking it)?

src/quantumchannel.jl
```diff
@@ -0,0 +1,86 @@
...
+      throw("Receiving register index already initialized")
```
Krastanov, line 81: I prefer to have `throw(exception_object)` or `error(string)`, but not `throw(string)`. Could you make the message a bit more detailed as well. E.g. something like "A `take!` operation is being performed on a `QuantumChannel` in order to swap the state into a `Register`, but the target register slot is not empty (it is already initialized) ."

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,109 @@
...
+queue = DelayQueue{Register}(sim, 10.0)
+
+qc = QuantumChannel(queue)
+
+@resumable function alice_node(env, qc)
+    put!(qc, regA[1])
+end
+
+@resumable function bob_node(env, qc)
+    @yield @process take!(env, qc, regB[1])
```
Krastanov, line 23: Why is the `@process` necessary here?

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,109 @@
...
+sim = Simulation()
+
+qc = QuantumChannel(sim, 10.0, T1Decay(0.1))
+
+@resumable function alice_node(env, qc)
+    put!(qc, regA[1])
+end
+
+@resumable function bob_node(env, qc)
+    @yield @process take!(env, qc, regB[1])
```
Krastanov, line 79: same question about `@process`

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,109 @@
...
+end
+
+@process alice_node(sim, qc)
+@process bob_node(sim, qc)
+
+run(sim)
```
Krastanov, line 85: Could you also check that the T1 decay process actually happens?

E.g. you can do `observable(regB[1:2], projector(bell))` which should give you a number smaller than 1 (the fidelity of the Bell state). That number should be the same as:
```
reg = Register(2,[T1Decay(0.1), nothing]) # You might need a more detailed constructor, I am not sure
initialize!(reg[1:2], bell)
uptotime!(reg[1], delay)
observable(reg[1:2], projector(bell))
```

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,109 @@
...
+sim = Simulation()
+
+qc = QuantumChannel(sim, 10.0, T2Dephasing(0.1))
+
+@resumable function alice_node(env, qc)
+    put!(qc, regA[1])
+end
+
+@resumable function bob_node(env, qc)
+    @yield @process take!(env, qc, regB[1])
```
Krastanov, line 103: same question about `@process`

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,109 @@
...
+end
+
+@process alice_node(sim, qc)
+@process bob_node(sim, qc)
+
+run(sim)
```
Krastanov, line 109: same request to test whether the background acts correctly

test/test_quantumchannel.jl
```diff
@@ -0,0 +1,109 @@
+using QuantumSavory
```
Krastanov, line 1: Could you also add a test that checks `@test_throws` for the `take!` operation?

src/quantumchannel.jl
```diff
@@ -0,0 +1,86 @@
...
+QuantumChannel(env::ConcurrentSim.Simulation, delay, background=nothing) = QuantumChannel(ConcurrentSim.DelayQueue{Register}(env, delay), background)
+
+function Base.put!(qc::QuantumChannel, rref::RegRef)
+  if !isnothing(qc.background)
+      uptotime!(rref, ConcurrentSim.now(qc.queue.store.env))
+  end
+
+  channel_reg = Register(1)
+  channel_reg.backgrounds[1] = qc.background
+  channel_reg.traits[1] = qc.trait
```
Krastanov, line 68: There is a `Register` constructor that takes the background and trait directly, no need to modify slots after creation. Could you use it to simplify the code a bit?

Come to think of it, we probably also want a `Register(::QuantumChannel)` constructor. Could you add that as well?

Krastanov:
Thank you for all of this! I left a bunch of comments in. Some of them are fairly nit-picky -- it is so that we can establish some coding style conventions that might slowly become actual APIs. My nit-pickiness will diminish in future pull requests.

Abhishek-1Bhatt:
Please  feel free to be as detailed with the reviews as you'd like, it can only help with improving the code quality of the repo, besides helping me code better ; )

test/test_quantumchannel.jl
Abhishek-1Bhatt: If I remove the `@ process` from here, with everything else remaining the same, it errors with:
```julia
ERROR: MethodError: no method matching state(::var"##Base.take!_FSMI#315"{QuantumChannel, Any, RegRef, Any})

Closest candidates are:
  state(::AbstractEvent)
   @ ConcurrentSim C:\Users\Abhishek Bhatt\.julia\dev\ConcurrentSim\src\base.jl:45

Stacktrace:
 [1] execute(ev::ConcurrentSim.Initialize, proc::Process)
   @ ConcurrentSim C:\Users\Abhishek Bhatt\.julia\dev\ConcurrentSim\src\processes.jl:37
 [2] (::ConcurrentSim.var"#1#2"{typeof(ConcurrentSim.execute), ConcurrentSim.Initialize, Tuple{Process}})()
   @ ConcurrentSim C:\Users\Abhishek Bhatt\.julia\dev\ConcurrentSim\src\base.jl:51
 [3] step(sim::Simulation)
   @ ConcurrentSim C:\Users\Abhishek Bhatt\.julia\dev\ConcurrentSim\src\simulations.jl:39
 [4] run(env::Simulation, until::ConcurrentSim.Timeout)
   @ ConcurrentSim C:\Users\Abhishek Bhatt\.julia\dev\ConcurrentSim\src\base.jl:91
 [5] run
   @ C:\Users\Abhishek Bhatt\.julia\dev\ConcurrentSim\src\events.jl:29 [inlined]
 [6] run(env::Simulation)
   @ ConcurrentSim C:\Users\Abhishek Bhatt\.julia\dev\ConcurrentSim\src\events.jl:29
 [7] top-level scope
   @ c:\Users\Abhishek Bhatt\.julia\dev\channels.jl:70
```

Abhishek-1Bhatt: and if I remove the `@yield` as well, doesn't error but skips any computation after the `@yield` in `take!`, not sure how to get around this behaviour

Abhishek-1Bhatt: It seems like the @process is necessary here because without the `QuantumSavory.take!` being a process, the yield in `QuantumSavory.take!` is not a part of the env and hence it doesn't wait on the `get` object returned by https://github.com/JuliaDynamics/ConcurrentSim.jl/blob/8591bff85edb62556be142f141dafa506e0f7fae/src/resources/delayed_stores.jl#L57

The take! method is kinda similar to the consumer process in this example: https://github.com/JuliaDynamics/ConcurrentSim.jl/blob/8591bff85edb62556be142f141dafa506e0f7fae/src/resources/delayed_stores.jl#L16-L22

Krastanov:
As you make changes, could you click "resolve" on the comments that are already dealt with. It makes it a bit easier to read through.

test/test_quantumchannel.jl
Krastanov: The issue is that we want to keep this `take!` consistent with the one for the other types of stores. In all the other cases `take!` only needs a `@yield`, presumably because it already returns a process type.

src/quantumchannel.jl
```diff
@@ -0,0 +1,85 @@
...
+
+  if !isnothing(qc.background)
+      uptotime!(channel_reg[1], qc.queue.delay)
+  end
+
+  put!(qc.queue, channel_reg)
+end
+
+
+@resumable function Base.take!(env, qc::QuantumChannel, rref_rec::RegRef)
```
Krastanov, line 78: Your current `take` is `@resumable` which means you need `@yield @process` instead of just `@yield`. That is not what the API is supposed to be. You already did a few experiments related to this and we have already discussed it in person. After struggling with it for a bit, I think I found the proper way to do it and it is very similar to how `latency` is implemented for the `ConcurrentSim.DelayQueue`.

Basically you need
```
@resumable function post_take(env, take_event)
    item = @yield take_event
    modify_item = -item
    return modify_item
end

function Base.take!(s::YourStoreType)
    inner_take_event = take!(s.q)
    @process post_take(s.q.store.env, inner_take_event)
end
```

Here is a very detailed full example:
```
using ConcurrentSim
using ResumableFunctions

##

struct C
    q::DelayQueue
end

C(env, d) = C(DelayQueue(env, d))

function Base.put!(s, i)
    @info "entering put! $i at $(now(s.q.store.env))"
    put!(s.q, i)
end

@resumable function post_take(env, take_event)
    item = @yield take_event
    modify_item = -item
    @info "post_take got $item at $(now(env)) and modified it into $modify_item"
    return modify_item
end

function Base.take!(s)
    @info "entering take! at $(now(s.q.store.env))"
    inner_take_event = take!(s.q)
    @process post_take(s.q.store.env, inner_take_event)
end

##

sim = Simulation()
c = C(sim, 100.0)

@resumable function producer(sim)
    @yield timeout(sim, 0.1)
    for i in 1:10
        @yield timeout(sim, 1)
        @info "putting $i at $(now(sim))"
        put!(c, i)
        @info "put $i"
    end
end

@resumable function consumer(sim)
    @yield timeout(sim, 20.1)
    for i in 1:10
        @info "taking attempt ($i) at $(now(sim))"
        event = take!(c)
        @info "take event $event made, now yielding"
        content = @yield event
        @info "take yield gives $content at $(now(sim))"
    end
end

@process producer(sim)
@process consumer(sim)

run(sim)
```

Abhishek-1Bhatt: This is really neat, thanks!

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+end
+
+
+@resumable function post_take(env, take_event, rref)
+  channel_reg = @yield take_event
+  QuantumSavory.swap!(channel_reg[1], rref)
+end
+
+function Base.take!(qc::QuantumChannel, rref::RegRef)
+  if isassigned(rref)
```
Krastanov, line 84: this check should happen after `@yield take_event`. It is fine if the register is not empty when we schedule the taking (although probably ill-advised) -- the check should be happening when the "taking" is happening, not when the "taking" is scheduled. I think the easiest solution is to move it just before the `swap!(channel_reg[1], rref)`.

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
+"""
+QuantumChannel for transmitting qubits as `RegRef`s from one register to another in a quantum protocol simulation, 
+with a channel delay and under the influence of a background process.
+The function `put!` is used to put the `RegRef` containing the qubit in the channel, which can then be received by
+the receiving register after a specified delay using the take! method in a synchronous way.
+
+```jldoctest
+julia> using QuantumSavory; using ResumableFunctions; using ConcurrentSim
+
+julia> bell = (Z1⊗Z1 + Z2⊗Z2)/sqrt(2.0); regA = Register(2); regB = Register(2); initialize!((regA[1], regB[2]), bell);
```
Krastanov, line 10: Can we simplify the example a bit? Just have Alice and Bob have only one slot, and just have alice start with a qubit in state Z1. That way this will also end up much shorter.

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+  background::Any
+end
+
+QuantumChannel(queue::ConcurrentSim.DelayQueue{Register}, background=nothing, trait=Qubit()) = QuantumChannel(trait, queue, background)
+
+QuantumChannel(env::ConcurrentSim.Simulation, delay, background=nothing, trait=Qubit()) = QuantumChannel(ConcurrentSim.DelayQueue{Register}(env, delay), background, trait)
+Register(qc::QuantumChannel) = Register([qc.trait], [qc.background])
+
+function Base.put!(qc::QuantumChannel, rref::RegRef)
+  if !isnothing(qc.background)
```
Krastanov, line 63: why does it matter whether qc has any background? This is about making sure `rref` is sufficiently updated. I think there should be no check here.

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+      uptotime!(rref, ConcurrentSim.now(qc.queue.store.env))
+  end
+
+  channel_reg = Register(qc)
+  swap!(rref, channel_reg[1])
+
+  if !isnothing(qc.background)
```
Krastanov, line 70: similarly, I do not think we want to have this check

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+      uptotime!(channel_reg[1], qc.queue.delay)
```
Krastanov, line 71: If `swap!` properly updates the `time` value, then this should be `qc.queue.delay+now(qc.queue.store.env)`

src/quantumchannel.jl
```diff
@@ -0,0 +1,89 @@
...
+QuantumChannel(env::ConcurrentSim.Simulation, delay, background=nothing, trait=Qubit()) = QuantumChannel(ConcurrentSim.DelayQueue{Register}(env, delay), background, trait)
+Register(qc::QuantumChannel) = Register([qc.trait], [qc.background])
+
+function Base.put!(qc::QuantumChannel, rref::RegRef)
+  if !isnothing(qc.background)
+      uptotime!(rref, ConcurrentSim.now(qc.queue.store.env))
+  end
+
+  channel_reg = Register(qc)
+  swap!(rref, channel_reg[1])
```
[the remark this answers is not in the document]
Abhishek-1Bhatt: Do you mean the `accesstimes` attribute? No, it only swaps the `staterefs` and `stateindices` of the `RegRef`s, should we edit the current swap! to do so?

Krastanov:
I was unhappy with how cumbersome it was to use `swap!` so I made a few improvements there.

I edited the docstring a bit: (1) to remove implementation details and focus on the "why" (2) I think one of the doctests would have failed because it contained an arbitrary memory address.

But (2) also showed me that some tests are not running, so I double checked a few things there... We have not been checking doctests for a while. Lets make buildkite a priority.

PR #55, opened 2023-11-01
Updated superdense coding circuit to work with `RegRef`s

Abhishek-1Bhatt:
Updated superdense coding circuits to work with `RegRef`s instead of assuming that the qubit is always at the first slot of the passed register


Krastanov:
Thank you! Merging!

PR #60, opened 2023-11-17
Fixed doctests

Abhishek-1Bhatt:


Krastanov:
Is buildkite not running because we do not have a queue entry in it anymore?

Krastanov:
nope that does not seem to be the reason...

Krastanov:
oh, I see! buildkite is not running on pull requests from third party repositories as a security precaution

image

... I enabled it... let's see how soon I will regret this

Krastanov:
ok, now I think all tests are running correctly.

There are two outstanding issues:

1. this test fails because of a change in the hash of some arbitrary object https://buildkite.com/quantumsavory/ci-buildkite/builds/56#018bdbbc-cbce-4e65-a5d9-5302adaf69c5/369-2581
I thought I had fixed that with the use of the filter above it, but I guess I must have done something incorrectly with the filter.
2. A lot of the changes you had made like `Qubit -> QuantumSavory.Qubit` now seem like have to be reverted. I have had this happen often to me, it has something to do with what modules are imported when the tests run, and I frequently get it wrong (something to do with testing from the REPL vs testing from runtests.jl due to difference in exactly what is imported). It is a pretty annoying one :(

If point 1 takes you more than 15 minutes to fix, just change the docstring so that it just does not print that information. No need to waste a ton of time on something so menial.

Krastanov:
woohoo! Thank you! These green checkmarks make me feel so much better. I will merge shortly.

Krastanov (commented):
edit: comment placed by accident

Abhishek-1Bhatt:
Yes! it is really validating, haha

PR #61, opened 2023-11-17
Update pipeline.yml with custom xvfb plugin

Krastanov:


Krastanov:
@Abhishek-1Bhatt , I created a new buildkite plugin at https://github.com/QuantumSavory/julia-xvfb-buildkite-plugin that makes sure that xvfb (the X11 virtual framebuffer) is used to launch julia. It is quite ugly and unpolished of a plugin, but it seems to do the job.

Abhishek-1Bhatt:
Great, this resolves our GLMakie issue on the headless server!

PR #62, opened 2023-11-17
initial queries and tags setup

Krastanov:
@Abhishek-1Bhatt , this is the unfinished queries and tags system. I will add a bit of documentation and test and will hand it off to you in a day or two. After that we should start creating a few protocols that use it (in particular, simplifying a lot of the already created protocols)


Krastanov:
I do not know how I screwed this up, but mixing using the VS code git interface and the git cli led to this being pushed directly to master. My apologies.

PR #69, opened 2023-11-19
[draft] Firstgenrepeaterv2

Krastanov:
making this just so I can run the tests


src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -0,0 +1,172 @@
...
+        @yield lock(a) & lock(b) # this yield is expected to return immediately
+        @yield timeout(prot.sim, prot.local_busy_time_pre)
+        @yield timeout(prot.sim, (rand(Geometric(prot.success_prob))+1) * prot.attempt_time)
+        initialize!((a,b), prot.pairstate; time=now(prot.sim))
+        @yield timeout(prot.sim, prot.local_busy_time_post)
+        tag!(a, :EntanglementCounterpart, prot.nodeB, b.idx)
+        tag!(b, :EntanglementCounterpart, prot.nodeA, a.idx)
+        unlock(a)
+        unlock(b)
+        rounds==-1 || (rounds -= 1)
```
Abhishek-1Bhatt, line 88: When `rounds==-1` , but the entanglement has been successfully established, how will the while loop terminate? Or when `rounds` is a finite number and entanglement was established in the first iteration, would it still keep executing for all rounds? Or is it controlled from the simulation in some way?

Krastanov: The loop would not terminate, instead the protocol will continue searching for available free slots (lines 72 and 73) and if there are no free slots it will wait a bit and try again (lines 74-78).

Related: it would be nice to implement some form of `querywait` process that waits for an event to happen before running the query (in order to remove the `@yield timeout` step and instead it would just have `@yield something_changed_so_now_you_can_try_the_query_again`).

Abhishek-1Bhatt: I see, so the entangler process is always running in the background and entangling whenever a free slot is available at both the nodes.
Earlier, my understanding was that we only want a single entangled pair between each of the registers, but instead we want as many as possible and as often as possible

Krastanov: if `rounds=1` it will indeed try to do it only once and then it will exit the while loop. The default of `-1` is the one that leads to infinitely trying (hopefully... if I have not done something silly)

Krastanov: oh, and I guess the source of confusion is whether we are talking about "rounds of attempting entanglement" or "rounds of successfully generating entanglement". Here it is meant as the latter.

Krastanov:
closed in favor of #81

PR #81, opened 2023-12-14
entanglement tracker and swapper added to ProtocolZoo, with significant improvements to the query and messaging systems

Abhishek-1Bhatt:


Krastanov:
Could you squish the last 3 commits so all the changes relative to the previous PR are easy to read?

test/test_entanglement_tracker.jl
```diff
@@ -0,0 +1,103 @@
...
+using QuantumSavory.ProtocolZoo: EntanglementCounterpart, EntanglementHistory, EntanglementUpdateX, EntanglementUpdateZ
+using Graphs
+using Test
+using Random
+
+using Logging
+logger = ConsoleLogger(Logging.Debug; meta_formatter=(args...)->(:black,"",""))
+global_logger(logger)
+# same but this time with an entanglement tracker
+n = rand(1:30)
```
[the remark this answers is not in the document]
Krastanov: A random number is fine. I would even put it inside of the loop, so it tests for 1000 random samples instead having all of the tests at the same n.

test/test_entanglement_tracker.jl
```diff
@@ -75,27 +75,29 @@ using Logging
 logger = ConsoleLogger(Logging.Debug; meta_formatter=(args...)->(:black,"",""))
 global_logger(logger)
 # same but this time with an entanglement tracker
-n = 6 # works at <6
+n = rand(1:30)
 for i in 1:1000
     println("new =======================\n\n")
```
Krastanov, line 80: remove the print

test/test_entanglement_tracker.jl
```diff
@@ -75,27 +75,29 @@ using Logging
 logger = ConsoleLogger(Logging.Debug; meta_formatter=(args...)->(:black,"",""))
 global_logger(logger)
```
Krastanov, line 76: comment these out so that we can easily access them if we are debugging, but they are not in the way when just having normal tests -- this would create MB of data in the logs otherwise

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -294,9 +297,9 @@ end
                 error("`EntanglementTracker` on node $(prot.node) received a message $(msg) that it does not know how to handle (due to the absence of corresponding `EntanglementCounterpart` or `EntanglementHistory` tags). This is a bug in the protocol and should not happen -- please report an issue at QuantumSavory's repository.")
             end
         end
-        #println("tracker @$(prot.node) starting message wait at ", now(prot.sim), " with mb: ", mb.buffer)
-        @yield wait(mb)
-        #println("tracker @$(prot.node) message wait ends at ", now(prot.sim))
+        @debug "EntanglementTracker @$(prot.node): Starting message wait at $(now(prot.sim)) with MessageBuffer containing: $(mb.buffer)"
+        @yield wait(mb) # TODO -- this needs to be redone so that if a message was received before we started the wait, it returns immediately (so that we can look at the message)
```
Krastanov, line 301: I guess the TODO can be removed now that we fixed the issue

Krastanov:
> Could you squish the last 3 commits so all the changes relative to the previous PR are easy to read?

I should have said "squash". And I misread the number of commits. Could you squash all the new commits into a single commit with a description of what was done? If we need some additional steps after this is done we will squash them too (or just use `commit --amend`)

test/test_messagebuffer.jl
```diff
@@ -0,0 +1,32 @@
...
+using Test
+
+net = RegisterNet([Register(3), Register(2), Register(3)])
+env = get_time_tracker(net);
+@resumable function receive_tags(env)
+    while true
+        mb = messagebuffer(net, 2)
+        @yield wait(mb)
+        msg = querydelete!(mb, :second_tag, ❓, ❓)
+        print("t=$(now(env)): query returns ")
```
Abhishek-1Bhatt, line 13: Can we comment this out as well? Currently this is flooding the buildkite log and making is super slow to finish the tests

Krastanov: oh, yeah, that is some old silly debug message. The code should definitely never use println as it can not be filtered. At most `@debug` should be used.

src/messagebuffer.jl
```diff
@@ -0,0 +1,64 @@
...
+                for waiter in keys(mb.waiters)
+                    unlock(waiter)
+                end
+            end
+        end
+    end
+end
+
+function MessageBuffer(net, node::Int, qs::Vector{NamedTuple{(:src,:channel), Tuple{Int, DelayQueue{T}}}}) where {T}
+    sim = get_time_tracker(net)
```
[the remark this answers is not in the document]
Krastanov: Let's still keep the possibilities of a single-node network -- the asserts and descriptive error messages and special cases should be in whatever the rest of the API is.

test/test_entanglement_tracker.jl
```diff
@@ -7,9 +7,9 @@ using QuantumSavory.ProtocolZoo: EntanglementCounterpart, EntanglementHistory, E
 using Graphs
 using Test
 
-using Logging
-logger = ConsoleLogger(Logging.Debug; meta_formatter=(args...)->(:black,"",""))
-global_logger(logger)
+# using Logging
+# logger = ConsoleLogger(Logging.Debug; meta_formatter=(args...)->(:black,"",""))
+# global_logger(logger)
```
Krastanov, line 12: keep these commented out so they do not pollute the logs

src/messagebuffer.jl
```diff
@@ -37,16 +45,21 @@ end
 end
 
 function MessageBuffer(net, node::Int, qs::Vector{NamedTuple{(:src,:channel), Tuple{Int, DelayQueue{T}}}}) where {T}
-    sim = qs[1].channel.store.env
+    sim = get_time_tracker(net)
     signal = IdDict{Resource,Resource}()
-    mb = MessageBuffer{T}(sim, net, node, Tuple{Int,T}[], signal)
+    no_wait = [0]
```
Krastanov, line 50: this should probably be a `Ref{Int}()`, not an array

PR #86, opened 2024-01-04
EntanglementConsumer

Abhishek-1Bhatt:
EntanglementConsumer protocol for network simulations. Currently it uses `query` and `untag` and we empty the qubits manually by zeroing the stateindices and setting the staterefs to nothing for the slots, just for initial testing


src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -131,7 +132,8 @@ end
         tag!(a, EntanglementCounterpart, prot.nodeB, b.idx)
         # tag local node b with EntanglementCounterpart remote_node_idx_a remote_slot_idx_a
         tag!(b, EntanglementCounterpart, prot.nodeA, a.idx)
-
+        
+        @debug "EntanglerProt between $(prot.nodeA) and $(prot.nodeB): Entangled .$(a.idx) and .$(b.idx)"
```
Abhishek-1Bhatt, line 136: Please let me know if these new debug statements seem excessive, I'll remove them : )

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -305,4 +307,66 @@ end
...
+        q2 = query2.slot
+        @debug "EntanglementConsumer between $(prot.nodeA) and $(prot.nodeB): queries successful, consuming entanglement"
+        untag!(q1, query1.tag)
+        untag!(q2, query2.tag)
+        # TODO do we need to add EntanglementHistory and should that be a different EntanglementHistory since the current one is specifically for SwapperProt
+        # TODO currently when calculating the observable we assume that EntanglerProt.pairstate is always (|00⟩ + |11⟩)/√2, make it more general for other states
+        ob1 = observable((q1, q2), Z⊗Z)
+        ob2 = observable((q1, q2), X⊗X)
+
+        prot.net[prot.nodeA].staterefs[q1.idx] = nothing
```
Krastanov, line 362: We should avoid doing such low-level operations in the protocols. This should be just a traceout operation.

Currently this probably breaks some invariants -- what happens if these two slots were entangled to some third slot -- it will lead to a ton of issues if someone tries to do something with that third slot.

src/ProtocolZoo/ProtocolZoo.jl
Krastanov: nah, we should not shy away from adding a ton of debug info

Abhishek-1Bhatt:
With the query specification as fifo, we don't have issues with outdated history tags and we can run Entangler and Swapper for infinite rounds. But a new issue comes out of this, where we have  deadlock in the network due to a leftward node (say, node 2) having all its slots entangled with a single or multiple nodes only in the right(3, 4,...). This way, no swap can occur since we need a node in the left and one in the right to do a swap. Neither is it able to generate new entanglement with nodes on left since its slots can't be freed from entanglement with right nodes due to no swaps.

Krastanov:
Tests seem to be failing -- could you fix that? It makes reviewing much faster if the tests pass/

If it is because of a ton of noise in the logs due to warnings from SumTypes, you can add the same addition you had in #92

Krastanov:
The FIFO/FILO changes look good. I have not looked into the consumer in detail yet.

src/queries.jl
```diff
@@ -75,14 +75,14 @@ julia> queryall(r, :symbol, ❓, >(5))
...
 
-
 """
 $TYPEDSIGNATURES
 
 A query function searching for the first slot in a register that has a given tag.
 
 Wildcards are supported (instances of `Wildcard` also available as the constants [`W`](@ref) or the emoji [`❓`](@ref) which can be entered as `\\:question:` in the REPL).
 Predicate functions are also supported (they have to be `Int`↦`Bool` functions).
+The order of query lookup can be specified in terms of FIFO or FILO and defaults to FIFO if not specified.
```
Krastanov, line 85: I think the default is FILO? Not FIFO? At least that is what the kwargs seem to say.

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -166,8 +170,19 @@ end
...
     while rounds != 0
-        a = findfreeslot(prot.net[prot.nodeA], randomize=prot.randomize)
-        b = findfreeslot(prot.net[prot.nodeB], randomize=prot.randomize)
+        freeA = sum([!isassigned(prot.net[prot.nodeA][i]) for i in 1:length(prot.net[prot.nodeA].staterefs)])
+        freeB = sum([!isassigned(prot.net[prot.nodeB][i]) for i in 1:length(prot.net[prot.nodeB].staterefs)])
+        isentangled = !isnothing(query(prot.net[prot.nodeA], EntanglementCounterpart, prot.nodeB, ❓;assigned=true))
+        margin = isentangled ? prot.margin : prot.hardmargin
+        if freeA >= prot.margin && freeB >= prot.margin
+            a = findfreeslot(prot.net[prot.nodeA]; randomize=prot.randomize)
+            b = findfreeslot(prot.net[prot.nodeB]; randomize=prot.randomize)
```
Krastanov, line 179: This seems a bit roundabout. `freeA` and `freeB` basically are a reimplementation of `findfreeslot`. We should have something a bit more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

Let's have it instead more in the following style:

```
isentangled = !isnothing(query...)
margin = isentangled ? ...
a = findfreeslots(...; ..., margin)
b = findfreeslots(...; ..., margin)
```

That was we avoid reimplementing findfreeslot and have this capability available to other protocols. I believe this is the important abstraction to make available.

Krastanov:
This looks great! I made a couple of minor changes, hopefully did not break anything. I will plan to merge shortly.

Krastanov:
Thanks for fixing the issues I introduced!

PR #87, opened 2024-01-05
set compat upper bound for ResumableFunctions

Abhishek-1Bhatt:
Due to bugs in macro parsing with the current release of ResumableFunctions, setting an upper bound on the compat until the issues are resolved


Krastanov:
Could you file issues about these bugs, so we can keep track of things? We would want to easily remove these restrictions in the near future. Right now I am not sure we actually know what the issue is.

Issue #85, opened 2023-12-30
entanglement tracker and swapper on a grid (or otherwise not using the default predicates)


Krastanov:
In #81 we now have a `SwapperProt`. It has a `nodeL` and `nodeR` properties which are then used by `query` to decide whether a pair of local slots are a valid target for a swap. It is done by `findswapablequbits`.

Currently, this is tested only for a long chain of repeaters, using the default "wildcard" values for these properties.

More complicated setups should be tested.

- [ ] step 0

First, we need to make `findswapablequbits` sufficiently general. We need to support custom predicates, not only the `<(node)` and `>(node)` ones supported currently. 

- [ ] step 1

Run the current set of tests but on a grid. For instance, we have an entanglement generator between all nearest neighbors on a square grid, but only a chain through the diagonal is running swaps. In such a case, we will need a predicates that are not simply `<(node)` and `>(node)` but rather 'the other node is above and to the left of `node`' and 'the other node is below and to the right of `node`'.

- [ ] step 2

Good animated examples added to the documentation.

PR #90, opened 2024-01-17
Entanglement simulation on a grid with custom predicates

Abhishek-1Bhatt:
Following #85, here, an entanglement simulation  is implemented on a grid that runs `EntanglerProt` and `SwapperProt` only on the diagonal

### DIFF
--- a/src/ProtocolZoo/ProtocolZoo.jl
+++ b/src/ProtocolZoo/ProtocolZoo.jl
@@ -215,13 +215,8 @@ end
 function findswapablequbits(net, node, pred_low, pred_high) # TODO parameterize the query predicates and the findmin/findmax
     reg = net[node]
     
-    if typeof(pred_low) == Wildcard
-        l = <(node)
-        h = >(node)
-    else
-        l(x) = pred_low(net, node, x)
-        h(x) = pred_high(net, node, x)
-    end
+    l = typeof(pred_low) == Wildcard ? <(node) : pred_low
+    h = typeof(pred_high) == Wildcard ? >(node) : pred_high
 
     low_nodes  = queryall(reg, EntanglementCounterpart, l, ❓; locked=false, assigned=true)
     high_nodes = queryall(reg, EntanglementCounterpart, h, ❓; locked=false, assigned=true)
--- a/src/ProtocolZoo/ProtocolZoo.jl
+++ b/src/ProtocolZoo/ProtocolZoo.jl
@@ -226,8 +226,8 @@ function findswapablequbits(net, node, pred_low, pred_high, choose_low, choose_h
     high_nodes = queryall(reg, EntanglementCounterpart, pred_high, ❓; locked=false, assigned=true)
 
     (isempty(low_nodes) || isempty(high_nodes)) && return nothing
-    _, il = choose_low(n->n.tag[2], low_nodes) # TODO make [2] into a nice named property
-    _, ih = choose_high(n->n.tag[2], high_nodes)
+    il = choose_low((n.tag[2] for n in low_nodes)) # TODO make [2] into a nice named property
+    ih = choose_high((n.tag[2] for n in high_nodes))
     return low_nodes[il], high_nodes[ih]
 end
 
--- a/src/ProtocolZoo/ProtocolZoo.jl
+++ b/src/ProtocolZoo/ProtocolZoo.jl
@@ -175,7 +175,7 @@ end
     rounds = prot.rounds
     while rounds != 0
         reg = prot.net[prot.node]
-        qubit_pair = findswapablequbits(prot.net,prot.node)
+        qubit_pair = findswapablequbits(prot.net,prot.node, prot.nodeL, prot.nodeR)
         if isnothing(qubit_pair)
             isnothing(prot.retry_lock_time) && error("We do not yet support waiting on register to make qubits available") # TODO
             @yield timeout(prot.sim, prot.retry_lock_time)
@@ -212,16 +212,24 @@ end
     end
 end
 
-function findswapablequbits(net,node) # TODO parameterize the query predicates and the findmin/findmax
+function findswapablequbits(net, node, pred_low, pred_high) # TODO parameterize the query predicates and the findmin/findmax
     reg = net[node]
+    
+    if typeof(pred_low) == Wildcard
+        l = <(node)
+        r = >(node)
+    else
+        l(x) = pred_low(net, node, x)
+        h(x) = pred_high(net, node, x)
+    end
 
-    leftnodes  = queryall(reg, EntanglementCounterpart, <(node), ❓; locked=false, assigned=true)
-    rightnodes = queryall(reg, EntanglementCounterpart, >(node), ❓; locked=false, assigned=true)
+    low_nodes  = queryall(reg, EntanglementCounterpart, l, ❓; locked=false, assigned=true)
+    high_nodes = queryall(reg, EntanglementCounterpart, h, ❓; locked=false, assigned=true)
 
-    (isempty(leftnodes) || isempty(rightnodes)) && return nothing
-    _, il = findmin(n->n.tag[2], leftnodes) # TODO make [2] into a nice named property
-    _, ir = findmax(n->n.tag[2], rightnodes)
-    return leftnodes[il], rightnodes[ir]
+    (isempty(low_nodes) || isempty(high_nodes)) && return nothing
+    _, il = findmin(n->n.tag[2], low_nodes) # TODO make [2] into a nice named property
+    _, ih = findmax(n->n.tag[2], high_nodes)
+    return low_nodes[il], high_nodes[ih]
 end
 
 
--- a/src/ProtocolZoo/ProtocolZoo.jl
+++ b/src/ProtocolZoo/ProtocolZoo.jl
@@ -138,6 +138,9 @@ end
     end
 end
 
+function random_index(arr)
+    return rand(keys(arr))
+end
 
 """
 $TYPEDEF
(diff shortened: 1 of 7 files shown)


Abhishek-1Bhatt:
Wow, heavy breakage from new SumTypes release

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -212,16 +212,19 @@ end
     end
 end
 
-function findswapablequbits(net,node) # TODO parameterize the query predicates and the findmin/findmax
+function findswapablequbits(net, node, pred_low, pred_high) # TODO parameterize the query predicates and the findmin/findmax
     reg = net[node]
+    
+    l = typeof(pred_low) == Wildcard ? <(node) : pred_low
+    h = typeof(pred_high) == Wildcard ? >(node) : pred_high
```
Abhishek-1Bhatt, line 219: Do we need the above two lines?

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -212,16 +212,19 @@ end
...
 
-    (isempty(leftnodes) || isempty(rightnodes)) && return nothing
-    _, il = findmin(n->n.tag[2], leftnodes) # TODO make [2] into a nice named property
-    _, ir = findmax(n->n.tag[2], rightnodes)
-    return leftnodes[il], rightnodes[ir]
+    (isempty(low_nodes) || isempty(high_nodes)) && return nothing
+    _, il = findmin(n->n.tag[2], low_nodes) # TODO make [2] into a nice named property
+    _, ih = findmax(n->n.tag[2], high_nodes)
+    return low_nodes[il], high_nodes[ih]
 end
```
Abhishek-1Bhatt, line 228: Is the change in naming scheme seem suitable, I was hoping to come up with something more general since we are considering direction in a 2d grid as well now, instead of just left and right

Krastanov: Yup, that is a good idea! Thanks!

test/test_entanglement_tracker.jl
```diff
@@ -114,3 +114,136 @@ for i in 1:30, n in 2:30
     @test observable((q1.slot, q2.slot), Z⊗Z) ≈ 1
     @test observable((q1.slot, q2.slot), X⊗X) ≈ 1
 end
+
```
Krastanov, line 117: let's make all of this a separate file -- I want to be able to easily run a single file as a tests (when working interactively)

Krastanov:
It seems this is completely independent from #86 , right? I should be able to merge it shortly

src/ProtocolZoo/ProtocolZoo.jl
Krastanov: I do not think we should have these. My guess is that you are trying to set some defaults here? But I do not think we should treat the behavior for a simple chain as the automatic default.

Related to this, make sure that the examples what use this code are updated to explicitly use the `>(node)` and `<(node)` predicates.

test/test_entanglement_tracker.jl
```diff
@@ -114,3 +114,136 @@ for i in 1:30, n in 2:30
...
+    for (src, dst) in diag_pairs
+        eprot = EntanglerProt(sim, net, src, dst; rounds=1, randomize=true)
+        @process eprot()
+    end
+
+    for i in 2:n-1
+        l(x) = top_left(net, diag_nodes[i], x)
+        h(x) = bottom_right(net, diag_nodes[i], x)
+        swapper = SwapperProt(sim, net, diag_nodes[i]; nodeL = l, nodeR = h, rounds = 1)
```
Krastanov, line 232: given the change of names you made, we should probably also rename these to `nodeH` instead of `nodeR`

Krastanov:
This looks pretty good. I left some minor comments.

An example based on this should be added to the examples folder as well.

Krastanov:
The findmin and findmax that were used in findswappablequbits do not make sense if we are not in a 1D chain, so they should also be parameterized. I think I did that correctly. Please review my changes, fix any issues that I have introduced to make the tests pass again and merge without waiting for me. The "Downgrade" is failing due to unrelated issues.

Abhishek-1Bhatt:
The `argmin` and `argmax` return the argument directly, whereas, we were indexing the result from `findmin` and `findmax` earlier. I have reverted that change as the `random_index` function also required indexing. Let me know if we should stick with `argmin` and `argmax`, I'll change the `random_index` to something like `random_arg`.

Krastanov:
> The `argmin` and `argmax` return the argument directly, whereas, we were indexing the result from `findmin` and `findmax` earlier.

```
julia> argmax([20,10,30])
3
```

It returns an index, not the argument itself. Am I getting something wrong?

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -212,15 +219,15 @@ end
...
     reg = net[node]
 
     low_nodes  = queryall(reg, EntanglementCounterpart, pred_low, ❓; locked=false, assigned=true)
     high_nodes = queryall(reg, EntanglementCounterpart, pred_high, ❓; locked=false, assigned=true)
 
     (isempty(low_nodes) || isempty(high_nodes)) && return nothing
-    _, il = findmin(n->n.tag[2], low_nodes) # TODO make [2] into a nice named property
-    _, ih = findmax(n->n.tag[2], high_nodes)
+    _, il = choose_low(n->n.tag[2], low_nodes) # TODO make [2] into a nice named property
+    _, ih = choose_high(n->n.tag[2], high_nodes)
```
Krastanov, line 230: oh, you are saying I should have removed the underscore. Indeed, the underscore should not be there.

Abhishek-1Bhatt: No, in case of a vector of `NamedTuple` the output is the `NamedTuple`, the above snippet takes in a simple vector.

```julia
julia> using QuantumSavory; using QuantumSavory.ProtocolZoo: EntanglementCounterpart;

julia> reg = Register(5);

julia> a = [(slot=reg[9], tag=Tag(EntanglementCounterpart, 4, 1)), (slot=reg[8], tag=Tag(EntanglementCounterpart, 2, 5)), (slot=reg[4], tag=Tag(EntanglementCounterpart, 3, 1)), (slot=reg[1], tag=Tag(EntanglementCounterpart, 3, 1))]
4-element Vector{@NamedTuple{slot::RegRef, tag::Tag}}:
 (slot = Slot 9, tag = Entangled to 4.1)
 (slot = Slot 8, tag = Entangled to 2.5)
 (slot = Slot 4, tag = Entangled to 3.1)
 (slot = Slot 1, tag = Entangled to 3.1)

julia> argmax(n->n.tag[2], a)
(slot = Slot 9, tag = Entangled to 4.1)
```

Krastanov: Ah, yes, it should be just `il = choose_low(low_nodes). Then it all works, right? `argmax` has way more signatures than I expected. With that change the random_index function should also work. Apologies for the only partial update.

Abhishek-1Bhatt: `argmax`(argmin) returns the tuple: (slot = Slot 9, tag = Entangled to 4.1)
we can't use it to index the array of NamedTuple, so it doesn't work like that, if we want to do argmax, we would have to return `il` and `ih` directly

Krastanov: `argmax((n->n.tag[2] for n in low_nodes))`

Start from scratch here: what is the point of this change -- it is to make it easy for a user to provide a chooser function that gives you the preferred node to select. You can not have such a function depend on the internals of your code, you can not expect the user to know of it. Apologies for the repeated typos though, those are on me.

Abhishek-1Bhatt: The change totally makes sense. The problem is just that it the way it was implemented before(using `argmax` or `argmin`), it didn't work with indexing.
The third comment suggests doing `il = choose_low(low_nodes)`, that throws an error:

```julia
ERROR: MethodError: no method matching isless(::RegRef, ::RegRef)

Closest candidates are:
  isless(!Matched::Missing, ::Any)
   @ Base missing.jl:87
  isless(::Any, !Matched::Missing)
   @ Base missing.jl:88
```

Krastanov: Check my last comment. Does this work: `argmax((n->n.tag[2] for n in low_nodes))` or something to that effect.

Krastanov: The documentation comment also helps with this:

```
the `nodeL` predicate can return many positive candidates; `chooseL` picks one of them (by index into the array of filtered `nodeL` results), defaults to a random pick `arr->rand(keys(arr))`; if you are working on a repeater chain a good choice is `argmin`, i.e. the node furthest to the "left"
```

Abhishek-1Bhatt: I see, thanks, I got really confused in the beginning, sorry about that

test/test_entanglement_tracker_grid.jl
```diff
@@ -0,0 +1,151 @@
...
+end
+
+##
+# Here we test entanglement tracker and swapper protocols on an arbitrary hardcoded path of long-range connection inside of what is otherwise a 2D grid
+# We do NOT test anything related to automatic routing on such a grid -- only the hardcoded path is tested
+##
+
+## Custom Predicates
+
+function top_left(net, node, x)
```
Krastanov, line 24: I am waiting on CI and will then merge. However, this is an O(n) check for something that should be O(1).

From a node index get its x y coordinates (currently you get only one of them).
Then do the same for the other node.

Then just compare them to verify that you are to the left (one comparison) and to the top (second comparison).

No need for a for loop.

Submit a PR with the fix and feel free to merge it. It should also make this function much simpler.

Krastanov:
check the commit I added for some small cleanup changes I made

PR #92, opened 2024-01-25
added Query Interface and some doc fixes

Abhishek-1Bhatt:


docs/src/query_interface.md
```diff
@@ -0,0 +1,120 @@
...
+    -`remote_slot`
+    It indicates the current entanglement status with a remote node's slot.
+
+- `EntanglementHistory`
+    - `remote_node`
+    - `remote_slot`
+    - `swap_remote_node`
+    - `swap_remote_slot`
+    - `swapped_local`
+    This tag is used to store the outdated entanglement information after a swap. It helps to direct incoming entanglement update messages to the right node after a swap.
```
Krastanov, line 24: Be a bit more specific. "It helps in situations when locally we have performed a swap, but we are now receiving a message from a distant node that does not know yet that the swap has occurred (thus the distant node might have outdated information about who is entangled to whom and we need to update that information)."

docs/src/query_interface.md
```diff
@@ -0,0 +1,120 @@
...
+    - `past_local_node`
+    - `past_local_slot`
+    - `past_remote_slot`
+    - `new_remote_node`
+    - `new_remote_slot`
+    - `correction`
+    This tag arrives as a message from a remote node to which the current node was entangled to updat the entanglement information and apply a `Z` correction after the remote node performs an entanglement swap.
+
+The tags are constructed using the `Tag` constructor
+#### Tag(tagsymbol::Symbol, tagvariants...)
```
Krastanov, line 45: This is only one way to construct a tag, and in particular the tags described above are NOT constructed this way.

docs/src/query_interface.md
```diff
@@ -0,0 +1,120 @@
...
+where `tagvariants` are the extra arguments required by the specific `tagsymbol`, for instance the `tag_types.SymbolIntInt` require two `Int` values. It supports the use of predicate functions (`Int -> Bool`) and [`Wildcard`](@ref) (❓) in place of the `tagvariants` which allows the user to perform queries for tags fulfilling certain criteria.
+
+## `tag!`
+Adds a `Tag` to the list of tags associated with a [`RegRef`](@ref) in a [`Register`](@ref)
+#### `tag!(ref::RegRef, tag::Tag)`
```
Krastanov, line 50: These manual listings of method signatures should be automated otherwise they will become outdated very quickly. It is important to NOT rely on humans to keep documentation and code in sync -- that never works.

docs/src/query_interface.md
```diff
@@ -0,0 +1,120 @@
...
+    This tag arrives as a message from a remote node to which the current node was entangled to updat the entanglement information and apply an `X` correction after the remote node performs an entanglement swap.
+
+- `EntanglementUpdateZ`
+    - `past_local_node`
+    - `past_local_slot`
+    - `past_remote_slot`
+    - `new_remote_node`
+    - `new_remote_slot`
+    - `correction`
+    This tag arrives as a message from a remote node to which the current node was entangled to updat the entanglement information and apply a `Z` correction after the remote node performs an entanglement swap.
```
Krastanov, line 42: Elsewhere I made a comment about manual listing of methods. But this listing of types should also be done in an automated fashion, otherwise it will not be kept up to date with the code. One of the easier ways to do this is to use DocStringExtensions.jl to automatically generate all this information in the docstrings (and any of this manual text should also be first and foremost in the docstring). Than here you can just place an autodoc directive.

docs/src/query_interface.md
```diff
@@ -0,0 +1,120 @@
+# Query Interface
+
+```@meta
+DocTestSetup =  quote
+    using QuantumSavory
+end
+```
```
Krastanov, line 7: There should be an introductory paragraph or two explaining what is the point of all of this and why people should care about it. What problems does it solve (mostly, composability of protocols where one protocol can simply rely on queries and tags, instead of having to know the internals of other protocols).

Krastanov (commented):
This is a pretty good start. I have some minor comments.

The biggest one would be to rely more on actual docstrings. It makes sense that each function might have a small introductory paragraph, but most of the documentation should be set up so that it can reuse the docstrings. That is not well done yet in the "register interface" page either, but we should avoid adding additional documentation debt.

Another big thing is to avoid listing methods manually if possible -- in some situations, it has to be done because the automatic listing is simply too noisy. Please double check that is the case here (it might well be, in which case disregard my comment about automatic listing)

I would suggest installing the Grammarly extension, as there are occasional prepositions or typos that can be detected by it.

Another thing though. This is the "dev" documentation, i.e. a thing aimed at people that might want to contributes to the internals of the software. Much more important is to prepare tutorials or how-tos that show users why they should care about these capabilities.

docs/src/query_interface.md
```diff
@@ -0,0 +1,120 @@
...
+
+
+## `Tag`
+Tags are identifiers which are used to represent classical information needed for a quantum information. The library allows the construction of custom tags following the format of one of the [`tag_types`](@ref) using the `Tag` constructor. The library implements the following tags for use in the networking protocols:
```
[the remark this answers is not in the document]
Abhishek-1Bhatt: Ah, meant to say `quantum network`, whoops

docs/src/query_interface.md
Abhishek-1Bhatt: I couldn't get that, isn't this the only way we create tags. Even in ProtocolZoo, we never explicity create instances of the tags described above, its always done through `tag!` which is listed below this part. Is there something I'm missing here?

Abhishek-1Bhatt:
Keep getting duplicate docstring warning: https://github.com/QuantumSavory/QuantumSavory.jl/actions/runs/7744890873/job/21119667800?pr=92#step:5:1383
in the log. It triggers the duplicate docstring warning due to same object being  put inside `@docs` template more than once even if the signatures are clearly different

docs/make.jl
```diff
@@ -21,34 +21,36 @@ function main()
     modules = [QuantumSavory],
     authors = "Stefan Krastanov",
     pages = [
-    "QuantumSavory.jl" => "index.md",
-    "Getting Started Manual" => "manual.md",
+    # "QuantumSavory.jl" => "index.md",
```
Abhishek-1Bhatt, line 24: Looks like I forgot to uncomment these, from the local build.

PR #106, opened 2024-02-16
grid entanglement with multiple paths

Abhishek-1Bhatt:
Builds on #90


src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -111,11 +111,13 @@ end
 @resumable function (prot::EntanglerProt)(;_prot::EntanglerProt=prot)
     prot = _prot # weird workaround for no support for `struct A a::Int end; @resumable function (fa::A) return fa.a end`; see https://github.com/JuliaDynamics/ResumableFunctions.jl/issues/77
     rounds = prot.rounds
+    round = 1
     while rounds != 0
         a = findfreeslot(prot.net[prot.nodeA], randomize=prot.randomize)
         b = findfreeslot(prot.net[prot.nodeB], randomize=prot.randomize)
         if isnothing(a) || isnothing(b)
             isnothing(prot.retry_lock_time) && error("We do not yet support waiting on register to make qubits available") # TODO
+            @debug "EntanglerProt between $(prot.nodeA) and $(prot.nodeB)|round $(round): Failed to find free slots. \n Got:\n \t $a \n \t $b \n retrying..."
```
Abhishek-1Bhatt, line 120: I found round information to be helpful, during debugging so added it here and in `SwapperProt`. Let me know if this doesn't look good.

test/test_entanglement_tracker_grid.jl
```diff
@@ -7,145 +7,188 @@ using QuantumSavory.ProtocolZoo: EntanglementCounterpart, EntanglementHistory, E
...
 
-    run(sim, 200)
+    run(sim, 100)
 
-    q1 = query(net[1], EntanglementCounterpart, diag_nodes[n], ❓)
-    q2 = query(net[diag_nodes[n]], EntanglementCounterpart, 1, ❓)
-    @test q1.tag[2] == diag_nodes[n]
+    q1 = query(net[1], EntanglementCounterpart, size(graph)[1], ❓)
+    # q2 = query(net[size(graph)[1]], EntanglementCounterpart, 1, ❓)
+    q2 = (slot=net[q1.tag[2]][q1.tag[3]], tag = net[q1.tag[2]].tags[q1.tag[3]][1])
```
Abhishek-1Bhatt, line 189: This is an ugly fix to the situation where both `query`s return qubits from different entangled pairs which leads to `observable` returning nothing and the tests failing, https://github.com/QuantumSavory/QuantumSavory.jl/blob/1a0b40d45b5cfefbc75c8ea317af7d95bd58be02/src/baseops/observable.jl#L11

src/ProtocolZoo/ProtocolZoo.jl
Krastanov: looks good

src/queries.jl
```diff
@@ -95,11 +95,12 @@ julia> query(r, Int, 4, <(7))
 
 See also: [`queryall`](@ref), [`tag!`](@ref), [`Wildcard`](@ref)
 """
-function query(reg::Register, tag::Tag, ::Val{allB}=Val{false}(); locked::Union{Nothing,Bool}=nothing, assigned::Union{Nothing,Bool}=nothing) where {allB}
-    find = allB ? findall : findfirst
+function query(reg::Register, tag::Tag, ::Val{allB}=Val{false}(), ::Val{fifo}=Val{true}(); locked::Union{Nothing,Bool}=nothing, assigned::Union{Nothing,Bool}=nothing) where {allB, fifo}
```
Krastanov, line 98: this should probably be called `filo`, not fifo. First-in-last-out is what `findlast` would give you.

test/test_entanglement_tracker_grid.jl
```diff
@@ -7,145 +7,193 @@ using QuantumSavory.ProtocolZoo: EntanglementCounterpart, EntanglementHistory, E
...
 using Test
 
-if isinteractive()
-    using Logging
-    logger = ConsoleLogger(Logging.Debug; meta_formatter=(args...)->(:black,"",""))
-    global_logger(logger)
-    println("Logger set to debug")
-end
-
 ##
```
Krastanov, line 10: I do not really follow what is being tested here. Could you describe in words what each section is doing?

src/queries.jl
```diff
@@ -259,15 +261,26 @@ for (tagsymbol, tagvariant) in pairs(tag_types)
...
+                else
+                    for tag in tags
+                        if isvariant(tag, ($(tagsymbol,))[1]) # a weird workaround for interpolating a symbol as a symbol
+                            (_nothingor(locked, islocked(slot)) && _nothingor(assigned, isassigned(slot))) || continue
+                            if _all($(nonwild_checks...)) && _all($(wild_checks...))
+                                allB ? push!(res, (;slot, tag)) : return (;slot, tag)
+                            end
                         end
                     end
                 end
```
Krastanov, line 286:
```suggestion
                for tag in (filo ? reverse(tags) : tags) 
                    if isvariant(tag, ($(tagsymbol,))[1]) # a weird workaround for interpolating a symbol as a symbol
                        (_nothingor(locked, islocked(slot)) && _nothingor(assigned, isassigned(slot))) || continue
                        if _all($(nonwild_checks...)) && _all($(wild_checks...))
                            allB ? push!(res, (;slot, tag)) : return (;slot, tag)
                        end
                    end
                end
```

This way we avoid dangerous code repetition.

Krastanov: and then we want to be careful to do `reverse(res)` in the case of `allB == true and filo == true` at the end.

Krastanov:
I should have more time to look into this tonight if you have a chance to address my comments.

src/queries.jl
Abhishek-1Bhatt: We wouldn't need to do `reverse(res)` here, because based on `filo` we would already be looking serially backwards or forwards. We could probably remove the conditional in the `for` statement above and just do `reverse(res)` that should also work

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -185,9 +187,11 @@ end
         # tag local node b with EntanglementCounterpart remote_node_idx_a remote_slot_idx_a
         tag!(b, EntanglementCounterpart, prot.nodeA, a.idx)
 
+        @debug "EntanglerProt between $(prot.nodeA) and $(prot.nodeB)| round $(round): Entangled .$(a.idx) and .$(b.idx)"
```
Krastanov, line 190:
```suggestion
        @debug "EntanglerProt between $(prot.nodeA) and $(prot.nodeB)|round $(round): Entangled .$(a.idx) and .$(b.idx)"
```

src/queries.jl
```diff
@@ -322,7 +324,7 @@ for (tagsymbol, tagvariant) in pairs(tag_types)
...
             end
-            allB ? res : nothing
+            allB ? res : nothing # do we need to reverse here?
```
Krastanov, line 327: do not leave things like this in comments, especially not comments that do not have TODO or XXX marked to them, otherwise we will have silent bugs. Stuff like that needs to be fixed before sending for review

test/test_entanglement_tracker_grid.jl
```diff
@@ -7,145 +7,235 @@ using QuantumSavory.ProtocolZoo: EntanglementCounterpart, EntanglementHistory, E
 using Graphs
 using Test
 
-if isinteractive()
-    using Logging
-    logger = ConsoleLogger(Logging.Debug; meta_formatter=(args...)->(:black,"",""))
-    global_logger(logger)
-    println("Logger set to debug")
-end
```
Krastanov, line 15: why is this removed?

test/test_tags_and_queries.jl
```diff
@@ -32,3 +33,52 @@ tag!(r[5], Int, 4, 5)
 
 @test querydelete!(r[2], :symbol1, 4, ❓) == Tag(:symbol1, 4, 5)
 @test querydelete!(r[2], :symbol1, 4, ❓) === nothing
+
+
+# tests for fifo and filo order queries
+# for RegRefs
+reg = Register(5)
+tag!(reg[3], EntanglementCounterpart, 2, 3)
```
Krastanov, line 41: it is important for tests to be easy to read and understand. Here you have a very random looking order of entries, so it is very difficult for me to verify that your tests are actually correct. It would be much easier to have something like

```
tag!(reg[3], EntanglementCounterpart, 1, 2)
tag!(reg[3], EntanglementCounterpart, 3, 4)
tag!(reg[3], EntanglementCounterpart, 5, 6)
...
```

With repetitions or inversions or whatever else is necessary to make this tractable

src/queries.jl
```diff
@@ -123,17 +125,28 @@ julia> query(r, Int, 4, <(7))
...
-    find = allB ? findall : findfirst
-    i = find(i -> _nothingor(locked, islocked(reg[i])) && _nothingor(assigned, isassigned(reg[i])) && tag ∈ reg.tags[i],
-                  1:length(reg))
-    if allB
-        return NamedTuple{(:slot, :tag), Tuple{RegRef, Tag}}[(slot=reg[i], tag=tag) for i in i]
-    else
-        isnothing(i) ? nothing : (;slot=reg[i], tag=tag)
-    end
+function query(reg::Register, tag::Tag, ::Val{allB}=Val{false}(); locked::Union{Nothing,Bool}=nothing, assigned::Union{Nothing,Bool}=nothing, filo::Bool=true) where {allB}
+    _query(reg, tag, Val{allB}(), Val{filo}(); locked=locked, assigned=assigned)
```
Krastanov, line 129: this indirection from a Bool keyword argument to a Val{Bool} argument, is necessary if we want the compiler to guarantee that separate machine code is compiled for each of the two values (true / false).

It is also a way to avoid code duplication (e.g. `queryall ` simply calls `query(...,Val(true))`)

It made sense when the function was simpler and it makes sense for private things.

But for public things like the `filo` keyword, it is cumbersome to use; Moreover, it is not simple anymore, so we should probably just split these methods, instead of having all these compile-time if/else statements (in Julia an if-else that depends only on information in the type system, will be decided during compilation and there will be no if/else actually happening when we run the program)

test/test_tags_and_queries.jl
```diff
@@ -32,3 +33,73 @@ tag!(r[5], Int, 4, 5)
...
+
+@test query(reg, EntanglementCounterpart, 2, 22) == (slot = reg[2], depth = 7, tag = Tag(EntanglementCounterpart,2,22))
+@test queryall(reg, EntanglementCounterpart, 2, 22) == [(slot = reg[2], depth = 7, tag = Tag(EntanglementCounterpart,2,22)), (slot = reg[2], depth = 2, tag = Tag(EntanglementCounterpart,2,22))]
+@test queryall(reg, Tag(EntanglementCounterpart, 2, 22)) == [(slot = reg[2], depth = 7, tag = Tag(EntanglementCounterpart,2,22)), (slot = reg[2], depth = 2, tag = Tag(EntanglementCounterpart,2,22))]
+@test queryall(reg, EntanglementCounterpart, 2, 22) == queryall(reg, EntanglementCounterpart, ==(2), ==(22)) == queryall(reg, Tag(EntanglementCounterpart, 2, 22))
+@test queryall(reg, Tag(EntanglementCounterpart, 2, 22); filo=false) == reverse([(slot = reg[2], depth = 7, tag = Tag(EntanglementCounterpart,2,22)), (slot = reg[2], depth = 2, tag = Tag(EntanglementCounterpart,2,22))])
+@test queryall(reg, EntanglementCounterpart, 2, 22; filo=false) == queryall(reg, EntanglementCounterpart, ==(2), ==(22); filo=false) == queryall(reg, Tag(EntanglementCounterpart, 2, 22); filo=false)
```
Krastanov, line 105: all of these were failing before I fixed the issue with `queryall(::Register, ...)`

test/test_entanglement_tracker_grid.jl
Krastanov: I added it back

Krastanov:
@Abhishek-1Bhatt , thank you, this is merged now!

Please check out the changes I made, in terms of consistency of interface or ease of reading of tests.

Please learn more about how rebasing works in git and github. It is crucially important to get more comfortable with it, as it is the only way to scalably work on projects like this: a follow-up PR can not wait on this to be finished, otherwise work will never be done. Rebasing takes care tracking code changes. Testing should always be so thorough that it takes care of worries about how changes affect other code (even if we were not using much rebasing).

Krastanov:
Oh, I also marked a todo in one of the test files -- currently the protocols are tested only for a small finite number of rounds. That can hide many bugs -- it is very important to have tests where the number of rounds is not constrained. You are probably already investigating this in your other PRs, but it is worth mentioning it here.

PR #107, opened 2024-02-21
Repeater grid example and tutorial

### LINKED ISSUE #101: Grid entanglement simulation example
An example based on #90 should be added to the docs

Abhishek-1Bhatt:
Resolves #101


examples/repeatergrid/repeatergrid.jl
```diff
@@ -0,0 +1,109 @@
+using QuantumSavory
```
Krastanov, line 1: add this file to the example test runner, so that we can make sure that it does not break in the future. All other examples are executed by the test runner

Krastanov:
I left some minor comments. After they are addressed this can be merged and the as followup this should be planned (and added to the issue tracker):

- [ ] this deadlocks currently (as seen on the new plots that include the entanglement tracker) - that should be solved by improving the control protocols
- [ ] the entanglement tracker needs a better interface for accessing the log (mentioned in todos)
- [ ] an interactive webgl example should be made as well (similarly to how the other examples have interactive webgl versions). The WGLMakie + JSServe (now renamed to Bonito) are enough for that

Krastanov:
also, now that there are references in the docs, they seem to be causing documentation compilations issues -- that should be fixed too

Abhishek-1Bhatt:
Is the `GLFW` error in the buildkite ci related to out buildkite server?

PR #108, opened 2024-02-21
update lower bounds of dependencies and enable downgrade CI job

Abhishek-1Bhatt:


.github/workflows/downgrade.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           version: ${{ matrix.version }}
       - uses: cjdoris/julia-downgrade-compat-action@v1
```
Abhishek-1Bhatt, line 24: @Krastanov it looks like this action was recently archived, see: https://github.com/cjdoris/julia-downgrade-compat-action

Shoud I changed it to:
```julia
julia-actions/julia-downgrade-compat-action@v1
```
I have added all the base libraries to be skipped but it still fails maybe this will help?

Abhishek-1Bhatt:
This shouldn't happen: https://github.com/QuantumSavory/QuantumSavory.jl/actions/runs/8008861646/job/21876235207?pr=108#step:6:60
Since both Clifford and Savory have a lower bound set to `0.3.3` for QuantumInterface

Abhishek-1Bhatt:
Looks like its checking `v0.8.0` for QuantumClifford which has lower bound set to `0.1.0` for QuantumInterface,
https://github.com/QuantumSavory/QuantumClifford.jl/blob/7da8d351320b73e3f9461fa68b65c60ec03da3ea/Project.toml#L46

so we can either lower the bound for QuantumInterface or increase the bound for QuantumClifford which is currently set to `0.8`

Krastanov:
I would suggest always going for increasing bounds, in this case increasing the one for QuantumClifford.

.github/workflows/downgrade.yml
Krastanov: yes

Krastanov: oh, I see you have already done so

Krastanov:
@Abhishek-1Bhatt , the change to Distributions's lower bound is a bit awkward -- I made it because 0.25.0 does not support newer versions of FillArrays, which I learnt by using github blame on their project.toml file. This is probably not the most elegant and efficient way to run the downstream CI job, but it is better than nothing and it is not a high priority to make it better.

PR #114, opened 2024-04-15
Store time of tagging

### LINKED ISSUE #111: save time of tagging and use it in FILO/FIFO considerations
With #106 merged we now support filo and fifo order for query results. But if we have multiple slots, the order will be by slot index, not by time of insertion. We need to provide time of tagging for this to work.

Abhishek-1Bhatt:
Resolves #111


Abhishek-1Bhatt:
@Krastanov , any idea what the `xvfb` startup issue is from(the documentation ci failure)? https://github.com/QuantumSavory/QuantumSavory.jl/actions/runs/8932976512/job/24537665495?pr=114#step:7:564

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -190,9 +190,9 @@ end
         @yield timeout(prot.sim, prot.local_busy_time_post)
 
         # tag local node a with EntanglementCounterpart remote_node_idx_b remote_slot_idx_b
-        tag!(a, EntanglementCounterpart, prot.nodeB, b.idx)
+        tag!(a, EntanglementCounterpart, prot.nodeB, b.idx; tag_time = now(prot.sim), id = guid())
```
Krastanov, line 193: I do not think this is a viable solution. We can not ask the user to add a `guid` manually everywhere. This should be happening automatically behind the scenes.

src/networks.jl
```diff
@@ -14,6 +14,7 @@ struct RegisterNet
 end
 
 function RegisterNet(graph::SimpleGraph, registers, vertex_metadata, edge_metadata, directed_edge_metadata)
+    glcnt[] = 0 # set the global counter of `guid`s to zero whenever a new network is initialized
```
Krastanov, line 17: someone might be modeling multiple networks. No need to set it to zero, to avoid clashes

src/ProtocolZoo/ProtocolZoo.jl
Krastanov: Same with `tag_time`. This and `id` maybe can optionally be overridden, but by default they should happen automatically.

One option for time is to do `now(get_time_tracker(a))` where `a` has internally a reference to the time tracker.

src/queries.jl
```diff
@@ -301,8 +299,8 @@ for (tagsymbol, tagvariant) in pairs(tag_types)
     args = (:a, :b, :c, :d, :e, :f, :g)[1:length(sig)]
     argssig = [:($a::$t) for (a,t) in zip(args, sig)]
 
-    eval(quote function tag!(ref::RegRef, $(argssig...))
-        tag!(ref, ($tagvariant)($(args...)))
+    eval(quote function tag!(ref::RegRef, $(argssig...); tag_time::Union{Float64, Nothing}=nothing, id::Union{Int128, Nothing}=nothing)
```
Krastanov, line 302: these can probably just be `;kwargs...` so that we do not need to edit this line each time we add new keyword arguments

test/test_entanglement_consumer.jl
```diff
@@ -13,6 +13,7 @@ end
 
 
 for i in 1:30, n in 3:30
+    @show i, n
```
Krastanov, line 16: don't forget to comment this out in the final tests

src/networks.jl
Krastanov: On the other hand, this seems pretty useful when it comes to tests... So maybe we should not have a global counter, rather a counter per network?

Krastanov (changes requested):
I left some comments in. This seems to introduce an API change that breaks current code and requires the user to run a lot of extra boilerplate (carrying around the `guid()` calls). `guid` should not be a user-visible thing.

Not sure about why the documentation build fails. Do not worry about it for now.

src/networks.jl
Abhishek-1Bhatt: Its straight forward to make the counter local to each network, but then taking in to account the previous comment about adding `id` and `time` automatically inside the `tag!`, the local `id` now would be a field of the network and we would need to pass the network to `tag!` in order to access it

Krastanov: A new private field in the network makes sense, but having to pass around additional arguments is definitely not workable. The goal is not just to make something possible, equally important is for the interface to be something usable and logical.

Most of the objects that are being tagged carry a reference to the more centralized objects, e.g. the way `get_time_tracker` works. Maybe the best solution is to have `get_time_tracker` provide the object tracking the ids (e.g. by having a new simulation type).

Or you can just stick to a global non-zeroed counter, while fixing the rest of the things. That is fairly straightforward to solve separately.

Abhishek-1Bhatt: I think making `net` a field of our own `Simulation` type would be the best way about it then, as you suggested above. For now lets work with a the non-zeroed global counter, I'll open an issue for it

Abhishek-1Bhatt:
Same `GLFW` error here as https://github.com/QuantumSavory/QuantumSavory.jl/pull/107#issuecomment-2093927326

Krastanov:
is this ready for review?

Abhishek-1Bhatt:
Yes, sorry my bad, I should've re-requested for review

Krastanov:
no worries! You might have said so in person, but it helps to have it mentioned here as well in case I forget

src/queries.jl
```diff
@@ -2,28 +2,28 @@
 
 Assign a tag to a slot in a register.
 
-It returns the list of all currently present tags for that register.
-
 See also: [`query`](@ref), [`untag!`](@ref)"""
-function tag!(ref::RegRef, tag::Tag)
-    push!(ref.reg.tags[ref.idx], tag)
+function tag!(ref::RegRef, tag::Tag; tag_time::Union{Float64, Nothing}=nothing, id::Union{Int128, Nothing}=nothing)
```
Krastanov, line 6: - is `tag_time` ever used by any code we currently have?
- is `id` ever used by any code we currently have?

If not, let's remove the options to have them as input arguments and simplify the body of this function

src/queries.jl
```diff
@@ -2,28 +2,28 @@
...
+    tag_time = isnothing(tag_time) ? now(get_time_tracker(ref)) : tag_time
+    id = isnothing(id) ? guid() : id
+    push!(ref.reg.guids, id)
+    ref.reg.tag_info[id] = (tag, ref.idx, tag_time) 
 end
 
-tag!(ref, tag) = tag!(ref, Tag(tag))
+tag!(ref, tag; tag_time::Union{Float64, Nothing}=nothing, id::Union{Int128, Nothing}=nothing) = tag!(ref,Tag(tag); tag_time=tag_time, id=id)
```
Krastanov, line 13: same question about `tag_time` and `id`

src/queries.jl
```diff
@@ -2,28 +2,28 @@
...
 
-It returns the list of all currently present tags for that register.
-
 See also: [`query`](@ref), [`tag!`](@ref)
 """
-function untag!(ref::RegRef, tag::Tag) # TODO rather slow implementation. See issue #74
-    tags = ref.reg.tags[ref.idx]
-    i = findfirst(==(tag), tags)
-    isnothing(i) ? throw(KeyError(tag)) : deleteat!(tags, i) # TODO make sure there is a clear error message
+function untag!(ref::RegRef, id::Int128) # TODO rather slow implementation. See issue #74
```
Krastanov, line 22: Is the TODO comment still valid? What is necessary to fix it?

test/test_tags_and_queries.jl
```diff
@@ -82,26 +84,26 @@ for i in 2:4
...
 
-@test query(reg, EntanglementCounterpart, 2, 22) == (slot = reg[2], depth = 7, tag = Tag(EntanglementCounterpart,2,22))
-@test queryall(reg, EntanglementCounterpart, 2, 22) == [(slot = reg[2], depth = 7, tag = Tag(EntanglementCounterpart,2,22)), (slot = reg[2], depth = 2, tag = Tag(EntanglementCounterpart,2,22))]
-@test queryall(reg, Tag(EntanglementCounterpart, 2, 22)) == [(slot = reg[2], depth = 7, tag = Tag(EntanglementCounterpart,2,22)), (slot = reg[2], depth = 2, tag = Tag(EntanglementCounterpart,2,22))]
```
Krastanov, line 102: `depth` is not returned anymore? That is probably a good decision, just wanted to make sure it is on purpose?

test/test_tags_and_queries.jl
```diff
@@ -82,26 +84,26 @@ for i in 2:4
...
-@test query(reg, EntanglementCounterpart, 1, ❓, Val(false); filo=false) == (slot = reg[2], depth = 1, tag = Tag(EntanglementCounterpart,1,12))
-@test query(reg, EntanglementCounterpart, 1, ❓, Val(false); filo=true) == (slot = reg[2], depth = 8, tag = Tag(EntanglementCounterpart,1,312))
+@test query(reg, EntanglementCounterpart, 1, ❓, Val(false); filo=false) == (slot = reg[2], id = 13, tag = Tag(EntanglementCounterpart,1,12))
+@test query(reg, EntanglementCounterpart, 1, ❓, Val(false); filo=true) == (slot = reg[4], id = 36, tag = Tag(EntanglementCounterpart,1,314))
 
-@test queryall(reg, EntanglementCounterpart, 1, ❓) == [(slot = reg[i], depth = d, tag = Tag(EntanglementCounterpart,1,j+i)) for i in 2:4 for (d,j) in ((8,310),(5,110),(1,10))]
-@test queryall(reg, EntanglementCounterpart, 1, ❓; filo=true) == [(slot = reg[i], depth = d, tag = Tag(EntanglementCounterpart,1,j+i)) for i in 2:4 for (d,j) in ((8,310),(5,110),(1,10))]
-@test queryall(reg, EntanglementCounterpart, 1, ❓; filo=false) == [(slot = reg[i], depth = d, tag = Tag(EntanglementCounterpart,1,j+i)) for i in 2:4 for (d,j) in reverse(((8,310),(5,110),(1,10)))]
```
Krastanov, line 98: This test is very difficult to read. Tests should be obviously correct, not copy-pasting things that we generate by the code we are testing. This test is also fragile -- it will have to be modified if we switch  how the id is represented.

I would suggest making a helper function in this file along the lines of `strip_id` so that you can do `strip_id(querryall(...)) == ...` in order to keep the test easy to read.

Krastanov: It will be even easier to read than previously becaise we will not need the depth tag anywhere either.

Basically, remove `id` from these tests so that the tests are not fragile.

Krastanov (changes requested):
This seems nearly ready.

I think we are still missing tests for the new functionality, though. Tests that previously would have failed, like requesting `filo` or `fifo` for a whole register where we have various slots of the register tagged in arbitrary order. Could you add such tests?

src/queries.jl
Abhishek-1Bhatt: nope, they're only there just to provide the option to the user to override the guid function, but since this is an internal implementation detail, it would make sense to remove these

src/queries.jl
Abhishek-1Bhatt: I would consider it resolved, considering we're using a dictionary instead of a vector of tags, so deleting a tag just need the key to be dropped(deleted) rather than iterating and comparing for equality through each element of a vector of tags. If you agree I'll remove the TODO

test/test_tags_and_queries.jl
Abhishek-1Bhatt: Yes, there's no need for it since we just need the key(guid) to the dictionary id for accesing the metadata

Abhishek-1Bhatt:
> This seems nearly ready.
> 
> I think we are still missing tests for the new functionality, though. Tests that previously would have failed, like requesting `filo` or `fifo` for a whole register where we have various slots of the register tagged in arbitrary order. Could you add such tests?


It looks to me like these tests that we already have are what you described?
https://github.com/Abhishek-1Bhatt/QuantumSavory.jl/blob/f606f1baa76236d48bcbdc1065f208ff6628e2df/test/test_tags_and_queries.jl#L74-L100

test/test_tags_and_queries.jl
Krastanov: these tests still look much more complicated than they used to

Krastanov:
I made some changes and polish. Please check each of the commits I have submitted in case something is off. Most of them have a clear description for why I added them. Those marked `fixup` are just there to fix silly mistakes I have made as I was working on this.

Krastanov:
woohoo, this is in!

PR #142, opened 2024-08-02
Refactor tests to use TestItems.jl

BKaperick:
Closes issue #141.

Hello, this is a first contribution, so I am fully receptive to any feedback you have.

I have a few comments/questions:
* `@testitem` blocks cannot be nested, and I don't know of any way to share set-up code between them, so in `test/test_circuitzoo_purification.jl` I just kept the `@testset` blocks, but all nested within one `@testitem`.  If someone has a recommendation for how to do this more properly I will update this.

I am still testing it locally, so I am opening this as a draft to indicate that I am actively working on this issue.  I will change its status to Open once I've validated the changes are working correctly.


test/test_circuitzoo_purification.jl
```diff
@@ -1,44 +1,25 @@
...
+    @test_throws ArgumentError PurifyExpedient()(r[1], r[2], r[3:2:21]...)
+    @test_throws ArgumentError PurifyStringent()(r[1], r[2], r[3:2:21]...)
 end
-@test_throws ArgumentError PurifyExpedient()(r[1], r[2], r[3:2:21]...)
-@test_throws ArgumentError PurifyStringent()(r[1], r[2], r[3:2:21]...)
-
 
-@testset "2to1" begin
+@testitem "Circuit Zoo Purification - 2to1" tags=[:circuitzoo_purification] begin
+    include("setup_circuitzoo_purification.jl")
```
BKaperick, line 21: > @testitem blocks cannot be nested, and I don't know of any way to share set-up code between them, so in test/test_circuitzoo_purification.jl I just kept the @testset blocks, but all nested within one @testitem. If someone has a recommendation for how to do this more properly I will update this.

I have this working by refactoring the setup code to a separate file.  This seems to be the solution endorsed in [this thread](https://discourse.julialang.org/t/how-does-the-testing-pane-work/115301/8):
> Yes, that feature is just not finished. For now, the only real option to share code between different `@testitems` is to put it into a separate file, and then include that file from each @testitem.

Krastanov:
Hi, Bryan! Thank you for starting this! Please check out this similar contribution made to QuantumClifford.jl -- following the same style would help me a lot in keeping things consistent across projects: https://github.com/QuantumSavory/QuantumClifford.jl/pull/329

QuantumSavory probably will be a more challenging case because there are a lot of ad-hoc examples tested in weird environments (faked opengl environment in CI for some of the plotting tests for instance). It will probably need some redoing of of tagging and filtering of tests.

test/test_circuitzoo_purification.jl
Krastanov: That makes sense! Thanks!

test/Project.toml
```diff
@@ -7,6 +7,7 @@ DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
+GLMakie = "e9467ef8-e4e7-5192-8a1a-b1aee30e663a"
```
BKaperick, line 10: @Krastanov I saw this was removed in #126 but I don't have the full context for the problem it was solving.

Alternatively, re-adding `using Pkg; Pkg.add("GLMakie")` lines to the plotting test `@testitems` should work as well I think, and is closer to what was already there.

Also, when running tests from VSCode, these tests pass, however when manually running them via `runtests.jl` it was complaining:
```
LoadError: Invalid attribute markerspace for plot type 
MakieCore.LineSegments{Tuple{Vector{GeometryBasics.Point{2, Float32}}}}.
```
have you seen that before ?

BKaperick:
The only remaining failure is the [`persistent_tasks` failure](https://juliatesting.github.io/Aqua.jl/dev/persistent_tasks/) from the Aqua tests coming from `QuantumSavory.jl`.

Comparing the github actions [run log](https://github.com/QuantumSavory/QuantumSavory.jl/actions/runs/10282670061/job/28454844557#step:6:1194) to [a recent run on `master`](https://github.com/QuantumSavory/QuantumSavory.jl/actions/runs/9755785352/job/26924900841#logs), I notice one new warning during precompilation:

```
│  WARNING: Method definition traceout!(QuantumOpticsBase.Operator{BL, BR, T} where T where BR where BL, Any) in module QuantumOpticsBase at /home/runner/.julia/packages/QuantumOpticsBase/QN1iB/src/operators_dense.jl:31 overwritten in module QuantumSavory at /home/runner/work/QuantumSavory.jl/QuantumSavory.jl/src/backends/quantumoptics/quantumoptics.jl:21.
│  ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
```

and one new warning during running tests:

```
┌ Warning: Module QuantumSavory with build ID ffffffff-ffff-ffff-0000-009cee9a6d73 is missing from the cache.
│ This may mean QuantumSavory [2de2e421-972c-4cb5-a0c3-999c85908079] does not support precompilation but is imported by a module that does.
```

Krastanov:
Thanks, @BKaperick ! Some of the errors you are getting are from some invasive work happening to remove piracies that is still in progress as it spans multiple packages (e.g. #143 ). It should get merged soon and then I will be able to rerun the tests here and provide a review. Apologies for the delay!

I am running a summer workshop this week (qnumerics.org) so I might be a bit slow to answer.

Krastanov:
Apologies for not warning you about the JuMP/GraphMatching removal -- that happened because they were causing issues with installation on non-linux machines.

BKaperick:
Hi @Krastanov  no problem at all about the GraphMatching removal; I saw the other PR.

Right now I have 37 issues identified by JET, which is more than the 33 allowed.  The last run on `master` [also failed](https://github.com/QuantumSavory/QuantumSavory.jl/actions/runs/10871756781) but with only 35 issues.

Doing a diff on the JET output, it looks like these are the two new issues, but looking at [src/CircuitZoo/CircuitZoo.jl#L150](https://github.com/QuantumSavory/QuantumSavory.jl/blob/master/src/CircuitZoo/CircuitZoo.jl#L150) I don't see why this PR would cause these new issues to arise:

```
> ┌  @ QuantumSavory.CircuitZoo /home/runner/work/QuantumSavory.jl/QuantumSavory.jl/src/CircuitZoo/CircuitZoo.jl:150
> │┌ indexed_iterate(I::Nothing, i::Int64) @ Base ./tuple.jl:95
> ││ no matching method found `iterate(::Nothing)`: x = iterate(I::Nothing)
> │└────────────────────
> │┌ indexed_iterate(I::Nothing, i::Int64, state::Int64) @ Base ./tuple.jl:100
> ││ no matching method found `iterate(::Nothing, ::Int64)`: x = iterate(I::Nothing, state::Int64)
> │└────────────────────
```

BKaperick:
After some investigation, @Krastanov this is what I've found:

Removing `Cbc` dependency, and the `MOI` from `JuMP` (specifically [removing these two lines](https://github.com/QuantumSavory/QuantumSavory.jl/blob/dfd8c08db3ee4224bcd213c198fb1be752a86502/src/ProtocolZoo/switches.jl#L10) and the removals of `Cbc` and `JuMP` from the `Project.toml`  in PR #153) added the two extra JET errors to `master`.

Changes to `src/ProtocolZoo/switches.jl `in #153 added the two extra JET errors to `master` because some [temporarily] dead code is referencing non-imported variables from `Cbc and `JuMP`.  This isn't really an issue, but I just commented out the dead code in this PR to remove those from the JET output.

The two additional JET issues I commented above are still mysterious to me, but caught when we re-add the line ignoring `JuMP.Containers` from the JET test, so I interpret that to mean it's somehow external to `QuantumSavory.jl` so it's okay to ignore.

Krastanov:
Thank you! Apologies for the annoyance with this. I should be able to review and merge this before the end of the week

Krastanov:
The documentation runner indeed has been somewhat unreliable. The other runners seem fine. I will restart the doc builder a few more times if it fails again. That issue is does not seem related to your work.

PR #145, opened 2024-08-11
Fix observable call due to something being now a keyword

Tortar:
I think this should solve https://buildkite.com/quantumsavory/quantumsavory/builds/465#01914295-bba4-45a9-9ce5-c1e56d8da52c but didn't test it so let's see if the test now passes


Tortar:
it works

Krastanov:
thanks for taking care of this!

Issue #130, opened 2024-07-09
Improve static analysis tests with JET [$200]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

# Project: "Improve static analysis tests with JET" [$200]

We use JET for static analysis of the codebase. **To claim this bounty, fix 2/3 of the reported warnings**. Some of them are false positives that can be removed with extra type annotation. Some of them are due to dependencies, in which case you can simply blacklist the dependency (as already done). See the `test_jet.jl` test file and the JET CI run for the current state of affairs.

**Required skills**: General Julia and JET skills.

**Reviewer**: Stefan Krastanov

**Duration**: 1 month

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

mawutory:
hi @Krastanov, please i would like to take this on as a first issue.

Krastanov:
Hi, @mawutory ! Do you have past experience with Julia and the JET static analysis tool? While this issue should be relatively easy for someone with experience, it might be a bit hard jumping in if you are a novice, and starting with some of the other bug bounties might be easier.

mawutory:
hello @Krastanov, thanks for the warm introduction. i do have quite some experience in python but i'm fairly new to Julia. however, i realized that Julia syntax is quite similar to that of python so i do not think the language will be a problem for me but i do concede that i need time to understand the codebase and high-level computing concepts in general.

please i'd appreciate your advice on how think i could get started. what are the issues that you think could be a good first issue?

Krastanov:
[...]

PR #151, opened 2024-09-11
Adjust to aviatesk/JET.jl#532

lgoettgens:
Some progress towards https://github.com/QuantumSavory/QuantumSavory.jl/issues/130.

JET used to report many problems with `Base.@kwdef` (see https://github.com/aviatesk/JET.jl/issues/487). https://github.com/aviatesk/JET.jl/pull/532 got rid of these reports, but the effect was not visible in this repo due to the custom `ReportPass` subtype. But this is no longer needed as well, as https://github.com/aviatesk/JET.jl/issues/477 has been fixed in the meantime (also via https://github.com/aviatesk/JET.jl/pull/532).
This change on its own already gets rid of many JET warnings. Since it is so many at once, I wanted to put this into its own PR.

- [n/a] The code is properly formatted and commented.
- [n/a] Substantial new functionality is documented within the docs.
- [n/a] All new functionality is tested.

### DIFF
--- a/test/test_jet.jl
+++ b/test/test_jet.jl
@@ -2,24 +2,7 @@ using Test
 using QuantumSavory, JET
 using DiffEqBase, Graphs, JumpProcesses, Makie, ResumableFunctions, ConcurrentSim, QuantumOptics, QuantumOpticsBase, QuantumClifford, Symbolics, WignerSymbols
 
-using JET: ReportPass, BasicPass, InferenceErrorReport, UncaughtExceptionReport
-
-# Custom report pass that ignores `UncaughtExceptionReport`
-# Too coarse currently, but it serves to ignore the various
-# "may throw" messages for runtime errors we raise on purpose
-# (mostly on malformed user input)
-struct MayThrowIsOk <: ReportPass end
-
-# ignores `UncaughtExceptionReport` analyzed by `JETAnalyzer`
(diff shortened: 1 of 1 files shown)


lgoettgens:
IMO this doesn't need a mention in the changelog. What do you think @Krastanov ?

Krastanov:
Indeed, but I am a bit confused, what is happening here?

On face value I thought the previous code was skipping one type of checks and keeping all other checks in place. And the new code is not skipping that check... So the new code should be raising more issues, not fewer... It seems I am wrong, but I do not understand where am I wrong.

lgoettgens:
> Indeed, but I am a bit confused, what is happening here?
> 
> On face value I thought the previous code was skipping one type of checks and keeping all other checks in place. And the new code is not skipping that check... So the new code should be raising more issues, not fewer... It seems I am wrong, but I do not understand where am I wrong.

The default pass of JET changed in https://github.com/aviatesk/JET.jl/pull/532 to skip two kinds of checks. See https://github.com/aviatesk/JET.jl/pull/532/files#diff-0f48511859153a7101b7dfa0d84b91ca974ed0b74701e8eb2d08acffd868c073R1414-R1416

Krastanov:
Oh, jeez, thanks for the heads up on this!

Krastanov:
sorry, somehow I missed the original description -- you had already gave an answer before I even asked. I should not be reviewing after teaching :D

lgoettgens:
No worries :)

PR #152, opened 2024-09-12
Fix many things reported by JET

lgoettgens:
Resolve https://github.com/QuantumSavory/QuantumSavory.jl/issues/130.

- [n/a] The code is properly formatted and commented.
- [n/a] Substantial new functionality is documented within the docs.
- [n/a] All new functionality is tested.

I was able to bring the number of reports down to 33 on julia 1.10. On julia 1.11.0-rc3, there are only 19 left. Many of the remaining things are due to some iterator collection magic in julia base, where I at some point gave up trying to understand it.

What will certainly help with the JET report and type stability in general, but what I see out of scope of this small project:

- Use some type for `time` and restrict many of the type signatures for it to `Union{Nothing, TimeType}`. It really does not matter if one uses Float64 or DateTime or whatever for it, but this would help the compiler in some places.
- Resumable functions currently don't allow the compiler to do small union splitting. Consider the following function
```julia
function foo(x::Union{Nothing, Int})
  if isnothing(x)
    error("x is nothing")
  end
  # x gets inferred as Int
  ...
end
```
If we now tag this function as resumable, `x` will still be referred as `Union{Nothing, Int}` there. For the places that came up I circumvented this by introducing a new variable that just has type `Int`.


src/ProtocolZoo/swapping.jl
```diff
@@ -66,12 +66,14 @@ end
...
-        qubit_pair = findswapablequbits(prot.net, prot.node, prot.nodeL, prot.nodeH, prot.chooseL, prot.chooseH; agelimit=prot.agelimit)
-        if isnothing(qubit_pair)
+        qubit_pair_ = findswapablequbits(prot.net, prot.node, prot.nodeL, prot.nodeH, prot.chooseL, prot.chooseH; agelimit=prot.agelimit)
+        if isnothing(qubit_pair_)
             isnothing(prot.retry_lock_time) && error("We do not yet support waiting on register to make qubits available") # TODO
             @yield timeout(prot.sim, prot.retry_lock_time)
             continue
         end
+        # The compiler is not smart enough to figure out that qubit_pair_ is not nothing, so we need to tell it explicitly. A new variable name is needed due to @resumable.
+        qubit_pair = qubit_pair_::NTuple{2, Base.NamedTuple{(:slot, :id, :tag), Base.Tuple{RegRef, Int128, Tag}}}
```
lgoettgens, line 76:
```suggestion
        qubit_pair = qubit_pair_::NTuple{2, Base.NamedTuple{(:slot, :id, :tag), Base.Tuple{RegRef, Int128, Tag}}} # TODO: replace by `NTuple{2, @NamedTuple{slot::RegRef, id::Int128, tag::Tag}}` once https://github.com/JuliaDynamics/ResumableFunctions.jl/issues/104 is resolved
```

Krastanov:
Thank you so much! Please send me an email at skrastanov@umass.edu with your name and email and a link to the bounty, so I can set up your reimbursement process with NumFOCUS (it might take a few weeks as we do them in batches and they are still setting everything up for their first batch).

PR #156, opened 2024-09-23
waiting on register modification events

Krastanov:


src/semaphore.jl
```diff
@@ -0,0 +1,29 @@
...
+    if s.nbwaiters[] > 0
+        unlock(s.lock)
+    end
+end
+
+function unlock(s::AsymmetricSemaphore)
+    if s.nbwaiters[] > 0
+        unlock(s.lock)
+    end
+end
```
Krastanov, line 29: The rest of the library does not automatically know about the existence of this file. In needs to be included from the main file `src/QuantumSavory.jl`. Given that it is a dependency for the content of `states_registers.jl` it should probably be included before it. E.g. here: https://github.com/QuantumSavory/QuantumSavory.jl/blob/46cb2f30b0e16167f12ca63082f11c6a8c0f3303/src/QuantumSavory.jl#L82

src/states_registers.jl
```diff
@@ -1,5 +1,58 @@
...
+end
+
+function getindex(vec::StateIndexVector, index::Int)
+    return vec.data[index]
+end
+
+function setindex!(vec::StateIndexVector, value::Int, index::Int)
+    vec.data[index] = value
+    unlock(vec.waiter)
+end
```
Krastanov, line 54: What is this for? Can it be removed?

src/states_registers.jl
```diff
@@ -1,5 +1,58 @@
...
+    if s.nbwaiters[] > 0
+        unlock(s.lock)
+    end
+end
+
+function unlock(s::AsymmetricSemaphore)
+    if s.nbwaiters[] > 0
+        unlock(s.lock)
+    end
+end
```
Krastanov, line 34: This seems to be repeating the content of `semaphore.jl`. I think my other comment about `include`-ing `semaphore.jl` will help clean this up.

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -208,9 +208,14 @@ end
         b_ = findfreeslot(prot.net[prot.nodeB]; randomize=prot.randomize, margin=margin)
 
         if isnothing(a_) || isnothing(b_)
-            isnothing(prot.retry_lock_time) && error("We do not yet support waiting on register to make qubits available") # TODO
-            @debug "EntanglerProt between $(prot.nodeA) and $(prot.nodeB)|round $(round): Failed to find free slots. \nGot:\n1. \t $a_ \n2.\t $b_ \n retrying..."
-            @yield timeout(prot.sim, prot.retry_lock_time)
+            if isnothing(prot.retry_lock_time)
+                @debug "EntanglerProt between $(prot.nodeA) and $(prot.nodeB)|round $(round): Failed to find free slots. \nGot:\n1. \t $a_ \n2.\t $b_ \n waiting..."
+                @yield lock(prot.nodeA.stateindices.waiter)
+                @yield lock(prot.nodeB.stateindices.waiter)
```
Krastanov, line 214: I am not sure, but I think this should be `@yield lock(...) | lock(...)` so that it runs whenever there is a change to either one of them. This would need more testing for correctness though.

Krastanov (commented):
Thanks, seems like a good start! Let's discuss the points raised above.

Krastanov:
closing in favor of #164

PR #157, opened 2024-09-23
Features for simulation with different controlplane architectures for arbitrary network topologies

ba2tro:


ba2tro:
closed in favor of #165 and #167

PR #161, opened 2024-10-10
convert zmeas and xmeas in swapper to Int explicitly

### LINKED ISSUE #158: `CliffordRepr()` incompatible with `ProtocolZoo.SwapperProt`
Qubits in Registers initialized with `CliffordRepr()` cause the swapper to throw error, due to the use of `CircuitZoo.LocalEntanglementSwap` which calls `project_traceout` , that call ends up at the `QuantumClifford` function: https://github.com/QuantumSavory/QuantumSavory.jl/blob/eba99c03857c0c9b5e5f003654e4359fe70add4c/src/backends/clifford/express.jl#L6 
This returns a `UInt64` that is used to initialize the `EntanglementUpdate` tags in the swapper [here](https://github.com/QuantumSavory/QuantumSavory.jl/blob/eba99c03857c0c9b5e5f003654e4359fe70add4c/src/ProtocolZoo/swapping.jl#L95) as zmeas and xmeas, but the Tags only take integer arguments.

```julia
using QuantumSavory

reg = QuantumSavory.Register(3, CliffordRepr())
initialize!(reg[1:2], QuantumSavory.StabilizerState("ZZ XX"))

projection = project_traceout!(reg.staterefs[1].state[], reg.stateindices[1], σˣ)
```
```julia
julia> typeof(projection[1])
UInt64
```

ba2tro:
Resolves #158

### DIFF
--- a/src/ProtocolZoo/swapping.jl
+++ b/src/ProtocolZoo/swapping.jl
@@ -92,12 +92,12 @@ end
         xmeas, zmeas = swapcircuit(q1, q2)
         # send from here to new entanglement counterpart:
         # tag with EntanglementUpdateX past_local_node, past_local_slot_idx past_remote_slot_idx new_remote_node, new_remote_slot, correction
-        msg1 = Tag(EntanglementUpdateX, prot.node, q1.idx, tag1[3], tag2[2], tag2[3], xmeas)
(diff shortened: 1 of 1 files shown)


Krastanov:
same comment as in the other PR -- changes like this definitely need tests, please make sure you add them in the future (it is mentioned in the default PR template when opening one)

PR #162, opened 2024-10-10
Make classical and quantum channel delays customizable when initializ…

ba2tro:


Krastanov:
Thanks! I will merge this shortly, after the tests pass. In general though, please have tests when new functionality is added -- this is definitely something that should have a test to make sure it is actually working.

Issue #133, opened 2024-07-09
Waiting on register events [$800]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

# Project: Waiting on register events [$800]

Many of the protocols implemented in this simulation library have the form:
```
while True:
    @yield timeout(fixed_wait_time)
    if check_whether_a_property_is_true()
        do_something_important()
    end
end
```
It would be much better if there is an option for them to be written as
```
while True:
    @yield waiting_for_register_or_slot_state_to_change()
    if check_whether_a_property_is_true()
        do_something_important()
    end
end
```
That can be relatively easily implemented by creating the necessary resource objects on which the protocols can listen (APIs provided by ConcurrentSim) and then hooking these features in the register API (e.g. `initialize`, `apply!`, `traceout!`, etc). We probably will want different types of "waiters" for different types of changes that might happen. This will be necessary also for changes related to tagging and untagging, or receiving messages over a channel. Some attention will need to be paid to making sure that all these new checks do not cause severe performance penalties to code that does not use them.

To win this bounty, such an API would need to be implemented, incorporated in some of the examples that already exist, enabled in already existing protocols, documented, and tested. There should also be a new example that compares the performance of a protocol that uses the `@yield timeout` versus one that uses the `@yield wait_for_register_change`.

**Required skills**: Understanding of ResumableFunctions.jl and ConcurrentSim.jl

**Reviewer**: Stefan Krastanov

**Duration**: 2 months

**Publication**: In the next 2 years we plan to release a paper in a selective journal about this software. Contributing to this issue would deserve a co-authorship status on such a paper (if the contributor so desires)

#### Payout procedure:

[...]

PR #164, opened 2024-10-30
Waiting on Register Events

hanakl:
This PR is for Project: Waiting on Register Events (issue #133). 

As @Krastanov suggested, a new struct `AsymmetricSemaphore` has been implemented. It serves as a resource object that allows multiple processes to wait for an update. More specifically, each register now has a field `tag_waiter`, which is locked when protocol conditions are unmet and unlocked upon any change in the register's tag. This mechanism is used in the following protocols: `EntanglerProt`, `EntanglementConsumer`, and `SwapperProt`.

Tests have also been added to validate both the semaphore functionality and the updated protocols.


Krastanov:
coverage issues are probably due to this: https://github.com/JuliaDynamics/ResumableFunctions.jl/issues/46

test/test_semaphore.jl
```diff
@@ -0,0 +1,148 @@
...
+]
+
+
+@test length(log) == length(expected_log)
+
+for i in 1:length(log)
+    @test log[i] == expected_log[i]
+end
+
+end
```
Krastanov, line 88: this `end` statement cuts out the rest of the tests and they are not included in the testitem, thus they are never executed

Krastanov: fixed

src/ProtocolZoo/cutoff.jl
```diff
@@ -72,6 +69,8 @@ end
 
             unlock(slot)
         end
-        @yield timeout(prot.sim, prot.period)
+        if !isnothing(prot.period)
+            @yield timeout(prot.sim, prot.period)
+        end
```
Krastanov, line 74: I do not think this is safe as currently done. Seems like it might lead to infinite repetition of the query. I think there should be some `else` with a wait on tag update.

Krastanov:
@hanakl , this looks great. I made some cosmetic changes. Two important things:

- one of the tests was not actually being executed because it was outside of the testitem environment; that should be fixed now
- The cutoff protocol might have a hidden bug right now. Could you make sure it has an appropriate else branch with waiting on tag update when it is given `nothing` for the wait period? Add tests to verify it is functioning in both cases (maybe a completely new test file) and feel free to merge without waiting for another review (ping me when it is merged)

Make sure to use "squash and merge" when you are merging and to clean the content of the commit message. This PR has a lot of small "fixup"-style commits that we do not want in the main history of the repository.

PR #170, opened 2024-11-21
Background Map for Plotting

hanakl:
This PR is for a feature to add a map as a background for plotting.
 
The package `GeoMakie` was used, and the extension `QuantumSavoryGeoMakie` was created. Additionally, error-handling messages for extensions were implemented, as well as making plotting functions support different imputs.

The new function `generate_map` outputs a Makie Axis, which integrates seamlessly with `registernetplot_axis` as before.

Example usage: 
```
using GLMakie
using GeoMakie
using QuantumSavory
GLMakie.activate!()

net = RegisterNet([Register(2),Register(3),Register(2),Register(5)])

# add some states, entangle a few slots, perform some gates
initialize!(net[1,1])
initialize!(net[2,3], X₁)
initialize!((net[3,1],net[4,2]), X₁⊗Z₂)
apply!((net[2,3],net[3,1]), CNOT)

# create the plot
ax = generate_map()
fig, ax, plt, obs = registernetplot_axis(ax,net, registercoords=[Point2f(-118, 34), Point2f(-71, 42), Point2f(-111, 34), Point2f(-96, 32)], state_linecolor=:black)
fig
```
This generates the following figure:

image


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## v0.5.1-dev
 
 - Add `classical_delay` and `quantum_delay` as keyword arguments to the `RegisterNet` constructor to set a default global network edge latency.
+- Implement a feature to include a map as a background for plotting (`generate_map`).
```
Krastanov, line 6: a minor point on technical writing style: "implement a feature to do x" is a bit wordy. "implement" already implies there is a "feature" and vice-versa. Something that more immediately gets to "do x" reads more naturally. You would notice though that I do not always remember to fix this style issue in my own writing

ext/QuantumSavoryMakie/QuantumSavoryMakie.jl
```diff
@@ -280,27 +280,44 @@ end
 
 ##
 
-"""Draw the given registers on a given Makie axis.
```
Krastanov, line 283: Does this not support subfigures anymore? That would be a breaking change. 

But reading through the rest of the code, it seems subfigures are still supported, so I guess it is just incorrect edit to the docstring.

Krastanov: nope, forget this comment, I just got confused

src/plots.jl
```diff
@@ -1,24 +1,59 @@
 """Draw the given register network.
 
 Requires a Makie backend be already imported."""
-function registernetplot end
+function registernetplot(args...; kwargs...)
```
Krastanov, line 4: This workaround of using `get_extension` is something I set up for use with types and structs, but it is not a very good way to deal with functions, given the multimethod nature of julia.
Methods of functions can be added in extensions -- that would be much more natural than what we are doing here, of using a new function in a new namespace.

For now we can probably keep this implementation, but it would be important to create an issue to track fixing this hack.

Potentially a better way to do this would be to have a hint attached to method errors. Something like what is done in this PR https://github.com/QuantumSavory/QuantumClifford.jl/pull/416/files# but just warning that the method error happened because the necessary package was not imported.

Krastanov: actually, it seems it is worse than that. Currently the way you have namespaced things it looks like a method is being overridden instead of a new function being made. That can be easily fixed by just restricting what is being imported in the namespace and keeping your current implementation.

But given that things need to be fixed anyway, I suggest doing the proper fix with method error hints that I described above.

Here is the errors I am currently getting

```
[ Info: Precompiling QuantumSavoryMakie [92a66160-645b-5231-a29a-c456122f462a] (cache misses: wrong dep version loaded (4), incompatible header (2))
WARNING: Method definition registernetplot(Any...) in module QuantumSavory at /home/stefan/Documents/ScratchSpace/quantumjulia/QuantumSavory.jl/src/plots.jl:4 overwritten in module QuantumSavoryMakie at /home/stefan/.julia/packages/MakieCore/EU17Y/src/recipes.jl:194.
ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
```

Krastanov:
what is the status of this?

hanakl:
@Krastanov It is waiting for your final review.

Krastanov:
ah! Looks good, please proceed with merging (it seems there is a minor merge conflict)

PR #174, opened 2024-12-10
fix: mermaid rendering in documenter-dark mode

Fe-r-oz:
This PR aims to provide the fix to mermaid charts in `register_interface.md `being unreadable in documenter-dark mode. 

For instance, 

Screenshot_select-area_20241210163346

I think this is due to the usage of `<code></code>` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code HTML tags that are used for in-line styling. But this causes rendering issues in the dark mode based on the observed evidence. The `<code>` tags are used to style text as inline code in HTML, and they are not required for mermaid diagrams to work.

```mermaid
flowchart TB
  A["uptotime!(ref::RegRef, now)"]
  B["uptotime!(refs::AbstractVecOrTuple{RegRef}, now)"]
  A --> B
```

Mermaid can automatically adjust to different themes, such as dark mode and light mode, by default color palette. For example, the diagram above appears as expected in dark mode.

Screenshot_select-area_20241210164111

### DIFF
--- a/docs/src/register_interface.md
+++ b/docs/src/register_interface.md
@@ -39,16 +39,16 @@ When a `state` is not provided, a default one is calculated from `newstate`, dep
 ```@raw html
 <div class="mermaid">
 flowchart TB
-  A["<code>initialize!(refs::Vector{RegRef}, state; time)</code>"]
-  B["<code>initialize!(r::Vector{Register}, i, state; time)</code>"]
+  A["initialize!(refs::Vector{RegRef}, state; time)"]
+  B["initialize!(r::Vector{Register}, i, state; time)"]
   subgraph TOP [lower from registers to states]
     direction LR
   end
-  D{{"<code>state<:Symbolic</code>"}}
+  D{{"state<:Symbolic"}}
   subgraph D1 [express state]
     direction LR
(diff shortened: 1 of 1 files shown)


Fe-r-oz:
In the [how to guide]( http://qs.quantumsavory.org/stable/howto/firstgenrepeater/firstgenrepeater/) for first gen repeater experiments, `<code>` tags are not used and the mermaids chart are displayed correctly. Therefore, in [register_inferace.md](https://github.com/QuantumSavory/QuantumSavory.jl/blob/master/docs/src/register_interface.md), we also replace the ` <code></code>` tags which is the purpose of this PR.

Please help review this PR, thank you!

Krastanov:
Thanks for submitting this @Fe-r-oz ! This is not a bad fix, but with it text that should be formatted as code is now formatted as prose, which I would like to avoid. On balance, if this was the only issue with mermaid diagrams in dark mode, I think I would have agreed it is a reasonable tradeoff. But after playing a bit with dark mode, it seems like mermaid diagrams are generally poorly rendered (e.g. with barely visible lines in the graph). This might be better solved by updates to the custom.css file used by the docs. Or upstream in DocumenterMermaid.jl.

PR #176, opened 2025-01-16
Move delayedProts out of the loop

hanakl:
This PR resolves method definition warnings that were cause by a test in [this PR ](https://github.com/QuantumSavory/QuantumSavory.jl/pull/164). The issue was that a resumable function (`delayedProts`) was defined inside a loop, which has now been fixed.


Krastanov:
LGTM if test failures are unrelated

PR #177, opened 2025-01-18
Avoid method-override warnings in tests by using separate modules

Krastanov:
By splitting the example tests in separate modules (each `@testitem` creates its own module) now we have separate namespaces and no worries about overwriting of methods.

I also removed a few other sources of spurious messages.

Lastly, I reorganized how the setup and tear-down for OpenGL plotting tests happens, hopefully leading to fewer crashes in buildkite.


Krastanov:
Let's see if the CI passes. If that is the case, could you do a quick review and merge if there is nothing worrisome, @hanakl ?

PR #184, opened 2025-03-01
Slot-based Entangling and Flexible Tag Consumption

hanakl:
This PR introduces two small adjustments to accommodate specific protocols under testing:

1. `EntanglerProt` now allows specifying slots to be entangled by either index or filter.
2. `EntanglerConsumer` can now consume any tag, defaulting to the original `EntanglementCounterpart`.


Krastanov:
closing and reopening to rerun tests

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -179,6 +179,10 @@ $TYPEDFIELDS
     rounds::Int = -1
     """maximum number of attempts to make per round (`-1` for infinite)"""
     attempts::Int = -1
+    """function or index of the slot to choose an available free slot in node A"""
+    chooseA::Union{Int,<:Function} = minimum
+    """function or index of the slot to choose an available free slot in node B"""
+    chooseB::Union{Int,<:Function} = minimum
```
Krastanov, line 185: I am thinking through comparisons of what is done in `SwapperProt` -- Just making sure this will not lead to a lot of confusion for someone who wants to use both.

I think you are right to not permit nodeA/nodeB to be wildcards, they are not like SwapperProt.nodeL/SwapperProt.nodeH, they are more like SwapperProt.node.

However chooseA sounds too much like SwapperProt.chooseL, but chooseA is for slots in a node while chooseL is for a node in a list of nodes.

Maybe we can call it `chooseAslot` and `chooseBslot`?

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -179,6 +179,10 @@ $TYPEDFIELDS
...
     """whether the protocol should find the first available free slots in the nodes to be entangled or check for free slots randomly from the available slots"""
     randomize::Bool = false
```
Krastanov, line 187: `randomize` and `chooseA` are now interdependent. I think it would be simpler to remove one from the struct and just use it in the constructor to potentially change what `chooseA` is.

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -204,8 +208,10 @@ end
     while rounds != 0
         isentangled = !isnothing(query(prot.net[prot.nodeA], EntanglementCounterpart, prot.nodeB, ❓; assigned=true))
         margin = isentangled ? prot.margin : prot.hardmargin
-        a_ = findfreeslot(prot.net[prot.nodeA]; randomize=prot.randomize, margin=margin)
-        b_ = findfreeslot(prot.net[prot.nodeB]; randomize=prot.randomize, margin=margin)
+        a_ = findfreeslot(prot.net[prot.nodeA]; filter=prot.chooseA, randomize=prot.randomize, margin=margin)
+        b_ = findfreeslot(prot.net[prot.nodeB]; filter=prot.chooseB, randomize=prot.randomize, margin=margin)
+        #a_ = findfreeslot(prot.net[prot.nodeA]; randomize=prot.randomize, margin=margin)
+        #b_ = findfreeslot(prot.net[prot.nodeB]; randomize=prot.randomize, margin=margin)
```
Krastanov, line 214: remove commented code

src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -389,6 +396,8 @@ $FIELDS
     nodeB::Int
     """time period between successive queries on the nodes (`nothing` for queuing up and waiting for available pairs)"""
     period::LT = 0.1
+    """tag for which the consumer is consuming"""
```
Krastanov, line 399:
```suggestion
    """tag type which the consumer is looking for -- the consumer query will be `query($node, EntanglementConsumer.tag, $remote_node)` and it will be expected that `$remote_node` possesses the symmetric reciprocal tag; defaults to `EntanglementCounterpart`"""
```

src/queries.jl
```diff
@@ -397,14 +397,30 @@ julia> findfreeslot(reg) |> isnothing
...
+        return nothing
+    end
+    freeslots = [i for i in 1:n_slots if !islocked(reg[i]) && !isassigned(reg[i])]
+    if isempty(freeslots)
+        return nothing
+    end
+    if filter isa Int
+        return filter in freeslots ? reg[filter] : nothing
+    else
+        filtered_slots = filter(freeslots)
```
Krastanov, line 413: this will raise an error if `freeslots` is the empty list

src/queries.jl
```diff
@@ -397,14 +397,30 @@ julia> findfreeslot(reg) |> isnothing
...
+        if filtered_slots === nothing || isempty(filtered_slots)
```
Krastanov, line 414: Why both checks? I prefer we are much more restrictive in what functiosn are permitted

src/queries.jl
```diff
@@ -397,14 +397,30 @@ julia> findfreeslot(reg) |> isnothing
...
+            return nothing
+        end       
+        if isa(filtered_slots, Integer)
```
Krastanov, line 417: same comment, we should be more restrictive

src/queries.jl
```diff
@@ -397,14 +397,30 @@ julia> findfreeslot(reg) |> isnothing
...
+            filtered_slots = [filtered_slots]
+        end
+        if randomize
```
Krastanov, line 420: I do not think randomize will actually provide a random result here as we already filtered the slots to be just the minimum slot

src/ProtocolZoo/ProtocolZoo.jl
Krastanov: maybe

```
    """function `Vector{Int}->Union{Nothing,Int}` or an integer slot number, specifying the slot to take among available free slots in node A"""
    chooseA::Union{Int,<:Function} = minimum
    """specifying the slot to take among available free slots in node A"""
    chooseB::Union{Int,<:Function} = minimum
```

src/ProtocolZoo/ProtocolZoo.jl
Krastanov: e.g. if the constructor is given `randomize=true`, then it overrides `chooseAslot` to be `rand`

src/queries.jl
```diff
@@ -397,14 +397,30 @@ julia> findfreeslot(reg) |> isnothing
...
-            slot = reg[i]
-            islocked(slot) || isassigned(slot) || return slot
+    n_freeslots = sum((!isassigned(reg[i]) for i in 1:n_slots))
+    if n_freeslots < margin
+        return nothing
+    end
+    freeslots = [i for i in 1:n_slots if !islocked(reg[i]) && !isassigned(reg[i])]
+    if isempty(freeslots)
+        return nothing
+    end
```
Krastanov, line 409:
```suggestion
    end
    if randomize
        filter = rand
    end
```

src/queries.jl
```diff
@@ -397,14 +397,30 @@ julia> findfreeslot(reg) |> isnothing
...
+        end       
+        if isa(filtered_slots, Integer)
+            filtered_slots = [filtered_slots]
+        end
+        if randomize
+            return reg[rand(filtered_slots)]
+        else
+            return reg[filtered_slots[1]]
         end
     end
```
Krastanov, line 425:
```suggestion
    if filter isa Int
        return filter in freeslots ? reg[filter] : nothing
    else
        filter_idx = filter(freeslots)
        return isnothing(filter_idx) ? nothing : reg[filter_idx]
    end
```

src/queries.jl
```diff
@@ -397,14 +397,30 @@ julia> findfreeslot(reg) |> isnothing
 true
 ```
 """
-function findfreeslot(reg::Register; randomize=false, margin=0)
+function findfreeslot(reg::Register; filter=minimum::Union{Int,<:Function}, randomize=false, margin=0)
```
Krastanov, line 400: let's rename `filter` to `choose` -- to me "filter" sounds like it might return more than one option

Krastanov (commented):
Thanks! I left in some suggestions for how to simplify this. Mostly, I prefer we do not leave as much "freedom" in how filtering can be done. There should be one permitted function signature, so that the API can be simpler.

Krastanov:
also, you can add this to your vscode configuration to make sure that whitespaces are being cleaned up, otherwise there is a lot of trailing whitespace noise that pollutes the git history:

```
    "files.trimFinalNewlines": true,
    "files.trimTrailingWhitespace": true,
    "files.eol": "\n",
```

Krastanov:
closing in favor of #192

PR #185, opened 2025-03-01
MBQC Purification

hanakl:
This PR showcases 2-to-1 purification using measurement-based quantum computing. Note that it uses the features in PR #184 .


Krastanov (changes requested):
it seems I have a pending review that was never submitted. Sorry!

PR #187, opened 2025-03-01
GHZ setup example

### LINKED ISSUE #202: simple first version of Emilly's quantum sensing over net
A first version of the protocol, together with an example, so that it is easy to review

see #187

hanakl:
This PR adds an example for setting up GHZ state for a quantum sensing protocol. Note that it depends on the changes made in #184 

Fixes #202


examples/GHZsetup/GHZsetup.jl
```diff
@@ -0,0 +1,92 @@
...
+
+        @yield timeout(sim, time)
+    end
+end
+
+
+# run the simulation
+
+
+net = RegisterNet(vcat([Register(1) for _ in 1:S], [Register(S)]))
```
Krastanov, line 71: There is a slightly neater syntax for concatenation:
```
julia> [[1,2,3]; 1]
4-element Vector{Int64}:
 1
 2
 3
 1
```

or in this case `RegisterNet([[Register... for _ in ...]; Register(S)])`

Now that I typed this out I am actually unsure which one is more legible. Pick whichever you like best, and consider this comment just an FYI about the other syntax.

examples/GHZsetup/GHZsetup.jl
```diff
@@ -0,0 +1,92 @@
...
+    @process eprot()
+end
+
+@process GHZ_projection(sim, net, S)
+run(sim, 10)
+
+ghz5 = StabilizerState("XXXXX ZZIII IZZII IIZZI IIIZZ")
+
+# a bit noisy, but should be close to 1
+observable([net[1], net[2], net[3], net[4], net[5]], [1, 1, 1, 1, 1], projector(ghz5))
```
Krastanov, line 92: does `net[1:S], ones(S)` work here? Or maybe `[net[i,1] for i in 1:S]`

examples/GHZsetup/GHZsetup.jl
```diff
@@ -0,0 +1,92 @@
...
+
+noisy_pair = noisy_pair_func(F)
+
+@resumable function GHZ_projection(sim, net, S; time=0.1)
+    while true
+        queries = []
+        incomplete = false
+        # check & wait for all entanglements
+        for i in 1:S
+            q = query(net[i], EntanglementCounterpart, S + 1, ❓; locked=false, assigned=true)
```
Krastanov, line 29: This query should probably be executed on `net[S+1]` and look for counterpart at `i`. Otherwise this implies instantaneous global knowledge of the network.

examples/GHZsetup/GHZsetup.jl
```diff
@@ -0,0 +1,92 @@
...
+    p = (4*F-1)/3
+    return noisy_pair_func_depol(p)
+end
+
+S = 5 # number of sensors
+F = 0.99 # fidelity
+
+noisy_pair = noisy_pair_func(F)
+
+@resumable function GHZ_projection(sim, net, S; time=0.1)
```
Krastanov, line 23: maybe adding a line here that says `hub_node = net[S]` would make the code a bit simpler

examples/GHZsetup/GHZsetup.jl
```diff
@@ -0,0 +1,92 @@
...
+        if m1 == 2
+            msg1 = Tag(EntanglementUpdateX, S + 1, 1, 1, -1, -1, m1)
+            put!(channel(net, S + 1 => 1; permit_forward=true), msg1)
+        end
+
+        for i in 2:S
+            m = project_traceout!(net[S + 1, i], Z)
+            msg = Tag(EntanglementUpdateZ, S + 1, i, 1, -1, -1, m)
+            put!(channel(net, S + 1 => i; permit_forward=true), msg)
+        end
```
Krastanov, line 61: Could you document for the reader why one node is receiving an X correction while the other nodes are receiving a Z correction -- just inline so someone reading the code can also learn the theory of the GHZ correction

examples/GHZsetup/GHZsetup.jl
```diff
@@ -0,0 +1,92 @@
...
+for i in 1:S
+    # entangle sensors with the hub
+    eprot = EntanglerProt(sim, net, i, S + 1; pairstate=noisy_pair, chooseA=1, chooseB=i, rounds=1, success_prob=1.)
+    @process eprot()
+end
+
+@process GHZ_projection(sim, net, S)
+run(sim, 10)
+
+ghz5 = StabilizerState("XXXXX ZZIII IZZII IIZZI IIIZZ")
```
Krastanov, line 89: can this be made dependent on S? There is `QuantumClifford.ghz(n)` that might work here, but it returns an internal to that library type. Hopefully `StabilizerState(QuantumClifford.ghz(S))` is enough?

Krastanov (changes requested):
Looks good!

- [ ] Could you add it to the test runner? -- there is a specific test file where examples can be imported
- [ ] I left a few stylistic comments, please address as you see fit
- [ ] and a few comments about inline documenting/comments
- [ ] may we rename it to something a bit more descriptive (presumably this will evolve into a complete implementation of Emily's protocol, at which point the name might be a bit unclear).  Maybe something like `GHZsensorhub` or `GHZ_sensor_network`

After these are addressed feel free to squash and merge without another review.

examples/GHZsetup/GHZsetup.jl
hanakl: Would it be better to explain the algorithm in more detail in the readme, or would adding a few comments here be sufficient?

Krastanov: ideally both, but inline for the code is what I feel is more important right now

PR #191, opened 2025-03-06
Update Map Plotting to Use Tyler instead of GeoMakie

hanakl:
We found that GeoMakie lacks the detailed city maps and labeling we needed, so we switched to Tyler. Additionally, we added a scaling factor to graph registers to accommodate different zoom levels of the map we will be using.

Here is an example of a map that can be generated.
Screenshot 2025-03-03 031445


Krastanov:
The example in `src/visualizations.md` seems to be outdated (causing the CI/Documentation failure)

Krastanov:
The failure in Downgrade is not from our package, rather from old versions of some dependencies (presumably because these dependency are not tested on the various linters (e.g. Aqua.jl) themselves). This looks good to merge!

PR #192, opened 2025-03-11
Slot-based Entangling and Flexible Tag Consumption

Krastanov:
Co-authored-by: hanakl

It seems #184 got quite messed up because of the sqash merge of #191? Here is a reset of it so it is easier to review:

Created by going to the most current master and checking out the changes of 184 without moving HEAD with the following command `git checkout hana_slot_choice -- .`

### DIFF
--- a/test/test_tags_and_queries.jl
+++ b/test/test_tags_and_queries.jl
@@ -150,7 +150,6 @@ id3 = tag!(reg[4], :symB, 4, 5)
 @test untag!(reg[1], id1).tag == Tag(:symA, 1, 2)
 @test untag!(reg, id2).tag == Tag(:symB, 2, 3)
 @test_throws "Attempted to delete a nonexistent" untag!(reg, -1)
-end
 
 ##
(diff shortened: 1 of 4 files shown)


Krastanov:
@hanakl , could you check that this PR fixes the issues with #184?

test/test_tags_and_queries.jl
```diff
@@ -150,6 +150,18 @@ id3 = tag!(reg[4], :symB, 4, 5)
...
 @test_throws "Attempted to delete a nonexistent" untag!(reg, -1)
+
+##
+# findfreeslot tests
+reg = Register(5)
+initialize!(reg[1], X)
+lock(reg[3])
+
+@test findfreeslot(reg).idx == 2
+@test findfreeslot(reg, filter=maximum).idx == 5
```
Krastanov, line 161: how come this works? This filter does not return a list

Krastanov:
The buildkite issues seems unrelated, although worrisome.

PR #200, opened 2025-03-23
redoing the ZALM source

Krastanov:
making it a bit more reproducible, with earlier source of truth


Krastanov:
it seems SciMLBase.jl and SciMLOperators.jl had some broken releases this afternoon that wrecked the whole ecosystem. Will rerun tests tomorrow, hopefully this is fixed by then.

Issue #206, opened 2025-03-27
Fix Tyler tile loading issue


hanakl:
The US map using Tyler.jl in the visualization section does not load properly.

PR #208, opened 2025-04-02
Tyler Map Fixes

hanakl:
This PR includes three small updates related to the Tyler mapping functionality, all based on suggestions from @Krastanov:

1. Fixed an issue where the Tyler map wasn't loading correctly in the documentation (see issue #206) by using the wait function.

2. Resolved a bug where registers would disappear when zooming by ensuring they are in the foreground.

3. Changed the default map to Positron for a cleaner look.

### DIFF
--- a/docs/src/visualizations.md
+++ b/docs/src/visualizations.md
@@ -51,8 +51,9 @@ initialize!(net[2,3], X₁) # hide
 initialize!((net[3,1],net[4,2]), X₁⊗Z₂) # hide
 apply!((net[2,3],net[3,1]), CNOT) # hide
 using Tyler
+coords = [Point2f(-118, 34), Point2f(-71, 42), Point2f(-111, 34), Point2f(-96, 32)]
 subfig, ax, map = generate_map()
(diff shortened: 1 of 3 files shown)


docs/src/visualizations.md
```diff
@@ -51,8 +51,9 @@ initialize!(net[2,3], X₁) # hide
 initialize!((net[3,1],net[4,2]), X₁⊗Z₂) # hide
 apply!((net[2,3],net[3,1]), CNOT) # hide
 using Tyler
+coords = [Point2f(-118, 34), Point2f(-71, 42), Point2f(-111, 34), Point2f(-96, 32)]
 subfig, ax, map = generate_map()
-fig, ax, plt, obs = registernetplot_axis(ax, net, registercoords=[Point2f(-118, 34), Point2f(-71, 42), Point2f(-111, 34), Point2f(-96, 32)])
+fig, ax, plt, obs = registernetplot_axis(ax, net, registercoords=coords, state_linecolor = :black)
```
hanakl, line 56: I can also update the default `state_linecolor`, since the current default gray color isn't very visible on the Positron map.

Krastanov (approved):
LGTM
Thanks!

PR #209, opened 2025-04-03
qtcp work

Krastanov:


src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -189,6 +194,8 @@ $TYPEDFIELDS
     margin::Int = 0
     """Like `margin`, but it is enforced even when no entanglement has been established yet. Usually smaller than `margin`."""
     hardmargin::Int = 0
+    """Tag to be added to the entangled qubits or nothing to not add any tag. The created tag will be of the form `tag(remote_node, remote_slot)`, by default `EntanglementCounterpart`."""
+    tag::Union{DataType,Nothing} = EntanglementCounterpart
```
Krastanov, line 198: @hanakl @Luisenden just flagging this recent change in case you are using EntanglerProt

Now you can specify what tag will be created, defaulting to `EntanglementCounterpart`. This is not a breaking change -- by default everything is still the same. But now you can change the tag if you wish.

PR #233, opened 2025-05-16
Add decoherence and subset GHZ projection

hanakl:


examples/GHZsetup/fixed_time.jl
```diff
@@ -0,0 +1,119 @@
...
+    procs = []
+    for i in 1:S
+        # entangle sensors with the hub (S+1)
+        eprot = EntanglerProt(sim, net, i, S + 1; pairstate=noisy_pair, chooseA=1, chooseB=i, success_prob=entanglemnt_success_prob)
+        @process eprot()
+    end
+    @yield timeout(sim, fixed_time)
+    for p in procs
+        cancel!(p)
+    end
```
Krastanov, line 94: - the `procs` is not populated
- I think it would be simpler to just do `EntanglerProt(...; success_prob=..., attempt_time=..., attempts=total_time/attempt_time, rounds=1)`

Krastanov: potentially add a comment that this works only if you have a large enough switch to not have waiting on a free slot

Krastanov (changes requested):
- much more documentation on what this is doing, in README for instance
- explanation of what the different files are
- explanation of how to get "scientific insight" out of these simulations
- in docs/readme/comments explanation of whether there is some global knowledge or instant/telepathic communication between nodes, or whether each node is working just with local information, generally explanation of what is being simulated
- reference to paper related to this
- structuring it similarly to other examples where there is "setup" and then there is "visualization" 
- adding an actual doc page, not just readme
- document that this is not the steady steady or that due to triviality this is immediately the steady state, e.g. because it always resets to the same initial configuration
- separate this into something like a setup, an example run that runs only one round, and something that visualizes statistics over many rounds or just repeated simulation over many rounds (specifically, see the simpleswitch example)

Check the readme of these examples for style and content guidelines:

- https://github.com/QuantumSavory/QuantumSavory.jl/tree/master/examples/congestionchain
- https://github.com/QuantumSavory/QuantumSavory.jl/tree/master/examples/simpleswitch
- https://github.com/QuantumSavory/QuantumSavory.jl/tree/master/examples/colorcentermodularcluster

After this is merged, consider making a simple wglmakie demo that we can run live as part of the docs.

Krastanov:
Let's rename the folder to one of the following:

- `ghz_networked_sensors`
- `ghz_sensor_distributed`
- `simple_networked_sensors`

or something along these lines, just a bit more descriptive.

And let's rename the files to be a bit more immediately obvious / standardized, following conventions from the other example folders.

Krastanov:
and add to the tests / examples files to make sure this is always executed during CI

PR #249, opened 2025-07-30
Add GUI Helper Functions

hanakl:
`available_slot_types()`, `available_background_types()`, and `constructor_metadata(::Type{<:AbstractBackgroundProcess})` were implemented to support the GUI project.  These are not part of the public API, and should be called with `QuantumSavory....`.  These functions are in `QuantumSavoryInteractiveUtils` extension, which requires the package `InteractiveUtils`.


Krastanov:
If tests pass I will go ahead and merge this and message the genie folks.

@hanakl , could you add the PrettyTable renderings of this to the documentation and move the InteractiveUtils dependencies to a package extension?

src/QuantumSavory.jl
```diff
@@ -58,7 +58,7 @@ using QuantumSymbolics: I # to avoid ambiguity with LinearAlgebra.I
 
 export
     StateRef, RegRef, Register,
-    Qubit, Qumode, QuantumStateTrait,
+    Qubit, Qumode, QuantumStateTrait, available_slot_types,
```
hanakl, line 61: I noticed that you didn't export the functions here, did you intend them to be called as QuantumSavory.function?

Krastanov: yeah, these should not be public API, otherwise we can not change them without making a breaking release

PR #250, opened 2025-08-04
Update documentation with new metadata functions

hanakl:
This PR adds documentation for the functions introduced in #249. It also includes the `PrettyTables` package in the documentation environment.


Krastanov:
The convention being "if it is mentioned in the documentation, it is public". I want to avoid that and keep these functions private, but still use them to make the docs better.

PR #252, opened 2025-08-16
Added a documentation file for integrating genqo with QuantumSavory

jgr-rgb:
Added a documentation file for integrating genqo (Python toolbox for modeling entanglement distribution sources using a hybrid Gaussian/non-Gaussian approach) with the StateZoo

### DIFF
--- a/src/StatesZoo/zalm/genqo_objects_documented.jl
+++ b/src/StatesZoo/zalm/genqo_objects_documented.jl
@@ -1,0 +1,239 @@
+"""
+Questions for Stefan:
+    - Is there an alternative to "MultiplexedCascadedBellPairW <: AbstractTwoQubitState" since we can also consider the photon-photon state?
+    - I have databases of states that were previously calculated. Is it possible for the visualization functions to use these as an alternative to calculating the state every time?
+"""
+
+using CondaPkg
+using PythonCall
+import LinearAlgebra: tr
+
+##
+
+CondaPkg.add_pip("genqo")
+
+##
+
+gq = pyimport("genqo")
+
+##
+
+"""
+$TYPEDEF
+
+Fields:
+
+$FIELDS
+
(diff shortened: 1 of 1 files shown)


Krastanov:
A few questions and requests:

- [ ] Check the changes I have pushed to this branch for whether they make sense to you. You can do that by clicking on the commit I have pushed above with title "minor cleanup" -- it will open a "diff" viewer that shows the changes just from that commit.
- [ ] It is considered polite to have library functions be quiet unless a progress bar is requested. Could you change the default in genqo to **not** print progress bars unless some configuration options requests them
- [ ] I tried ZALM with `ηᵇ, ηᵈ, ηᵗ, N, Pᵈ = 1,1,1,1,0.0001` and it took 7 minutes to get a single density operator on a very fast workstation; Is that expected/normal? If it is expected, what is the source of the slowdown, i.e. what is the one main expensive operation that leads to this?
- [ ] Could you make the genqo repository public? And may we move it to a github organization, to avoid the danger of it getting lost after you graduate? A typical setup would be to put it in a github org in which you can retain admin rights over the repository. We can put it in the QuantumSavory org for instance.
- [ ] Could you update the comments into actual docstrings? E.g. `# This function calculates A B C` should turn into `"""Calculate A B C"""` before the function header (not under it like Python does). No need to say "This function ..." just directly say what it does. There are one or two comments that also seem slightly mismatched (e.g. the fidelity function has a comment that says a density matrix is being calculated)
- [ ] For your SPDC source you have a reference. Could you add it to the bib file in the `/docs` folder and use `[...](@cite)` in the docstring (the way it is already done for the prajit and kevin references in the zalm one). That way the references renders properly in the autogenerated docs.

Before you do these changes make sure to first pull this branch to your local environment so that you have the changes I have pushed. You can verify it has worked by looking at `git log` -- it should list my commit as the top.

Krastanov:
the documentation failure is unrelated, due to an upstream change in PrettyTables

PR #257, opened 2025-08-22
introspection for available protocols

Krastanov:


Krastanov:
@hanakl , another one that can be used for an autogenerated table in the documentation

PR #266, opened 2025-09-04
use ASemaphore in messagebuffer as well

### LINKED ISSUE #229: the `onchange_tag(::Register)` and `wait(::MessageBuffer)` have the same functionality implemented in two different ways
`onchange_tag` uses `AsymetricSemaphore` while `wait` uses a dictionary of newly allocated resources in `MessageBuffer`.

They have a variety of consistency checks and tests, especially around bugs like #210 

We should

- make sure that both are tested for various edge cases -- currently their test are separate and it is possible that only one or the other is tested on particularly nasty edge cases
- benchmark both to see which one performs better (and which is typestable or non-allocating)
- remove one of the implementations and stick to the same implementation for both of them

Krastanov:
fixes #229


Krastanov:
well, the newest benchmarks are not encouraging... we will have to split this in separate parts as some things here are definitely useful, but it seems not all

PR #285, opened 2025-10-07
Implement electron-nuclear spin register for MBQC distillation protocol

hanakl:
For issue #204


examples/purificationMBQC/electron_nuclear_spin_register.jl
```diff
@@ -0,0 +1,235 @@
...
+
+
+# TODO: hardcoded for now
+function graph_generator(initial_entanglements_nodes, purified_nodes)
+    g = Graph(4)
+    for ij in [(1,3), (2,3), (3,4)]
+        add_edge!(g, ij...)
+    end
+    return g
+end
```
Krastanov, line 53: let's rename this to `purification_resource_graph(code)` or something like this

examples/purificationMBQC/electron_nuclear_spin_register.jl
```diff
@@ -0,0 +1,235 @@
...
+
+# TODO: right now, it just measures all but purified nodes
+@resumable function measure(sim, net, side, purified_nodes, storage_slot)
```
Krastanov, line 56: let's avoid assuming that this will be the only protocol running (e.g. the way you do index arithmetics with n depends on the assumption you have only two groups of nodes, only dedicated to running this protocol)

maybe;

`function measure(sim, net, local_storage_nodes_idx, storage_slot, remote_chief_node_idx)`

examples/purificationMBQC/electron_nuclear_spin_register.jl
```diff
@@ -0,0 +1,235 @@
...
+        msg = querydelete!(mb, XORMeasurements, ❓, ❓)
+        local_xor = local_tag.tag.data[3] # it would be better if it can be local_tag.tag.measurement
+        src, (_, src_node, remote_xor) = msg
+
+        if remote_xor == local_xor
+            @debug "Purification was successful"
+            tag!(local_tag.slot, PurifiedEntalgementCounterpart, src_node, storage_slot)
+        else
+            @debug "Purification failed."
+            untag!(local_tag.slot, local_tag.id)
```
Krastanov, line 120: and delete that qubit

examples/purificationMBQC/paper_implementation.jl
```diff
@@ -0,0 +1,447 @@
...
+        @yield reduce(&, (entanglers..., r1, r2))
+
+        println("resource & entangle ", now(sim))
+        @yield timeout(sim, 10)
+
+        m1 = @process alice_bell_meas()
+        m2 = @process bob_bell_meas()
+        @yield m1 & m2
+
+        println("measurements ", now(sim))
```
Krastanov: stuff like this, make sure it is `@info` so we can silence it during tests

examples/purificationMBQC/full_purification_example.jl
```diff
@@ -0,0 +1,171 @@
...
+    run(sim, 5.0)
+    for i in 1:k
+        alice_tag = query(net[alice_nodes[n+i]][storage_slot], PurifiedEntanglementCounterpart, ❓, ❓)
+        bob_tag = query(net[bob_nodes[n+i]][storage_slot], PurifiedEntanglementCounterpart, ❓, ❓)
+        # check whether purification was successful
+        @assert !isnothing(alice_tag)
+        @assert !isnothing(bob_tag)
+        f = observable([net[alice_nodes[n+i]], net[bob_nodes[n+i]]], [storage_slot, storage_slot], projector(perfect_pair))
+        @info "Purified pair $i: fidelity=$(round(f, digits=4)), alice_tag=$(alice_tag), bob_tag=$(bob_tag)"
+        @assert isapprox(f, 1.0) # Purified pair should be perfect
```
Krastanov: explain how this assert is a valuable correctness check

examples/purificationMBQC/full_purification_example.jl
```diff
@@ -0,0 +1,171 @@
...
+        sim_trial = get_time_tracker(net_trial)
+        @process run_purification(sim_trial, net_trial, n, resource_state, alice_nodes, bob_nodes,
+            communication_slot, storage_slot, pairstate_noisy, H1, H2, logxs, logzs, rounds=1)
+        run(sim_trial, 5.0)
+        tags = [query(net_trial[alice_nodes[n+i]][storage_slot], PurifiedEntanglementCounterpart, ❓, ❓) for i in 1:k]
+        if all(!isnothing, tags)
+            n_success += 1
+            fidelities = [observable([net_trial[alice_nodes[n+i]], net_trial[bob_nodes[n+i]]], [storage_slot, storage_slot], projector(perfect_pair)) for i in 1:k]
+            total_fidelity += sum(fidelities)
+            @assert all(f -> test_fidelity < f, fidelities)
```
Krastanov: explain in a comment why this is a good check and what it does

examples/purificationMBQC/full_purification_example.jl
```diff
@@ -0,0 +1,171 @@
...
+storage_slot = 2
+alice_nodes = collect(1:n+k)
+bob_nodes = collect(n+k+1:2*(n+k))
+
+# Perfect Bell Pairs: purification should be successful every time
+@info "Input fidelity: 1 (Perfect pair)"
+fidelity = 1
+pairstate = noisy_pair_func(perfect_pair, fidelity)
+for trial in 1:5
+    println(trial)
```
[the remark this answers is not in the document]
Krastanov: and a full sentence, not just a number

Krastanov (changes requested):
Some small requests for more inline docs

Could you post the log output of full_purification_example.jl before you merge, just so we see what it looks like

Krastanov (changes requested):
a few minor changes:

- [x] tests merge conflict
- [x] make a plotting example that just imports full_purification_example and makes two plots -- of success probability vs input F and of output F vs input F
- [x] the example should take just a minute
- [x] both examples should be tested, even if they are repetitive
- [x] do not export the content of the MBQC module, export only the module name itself

Krastanov (commented):
some test failures, but otherwise it looks good

```
#at the top of runtests.jl do
ENV["QS_TESTRUN"] = "true"

# in example file do
samples = get(ENV, "QS_TESTRUN")=="true" ? many : few
```

Krastanov-agent:
I traced the current JET failure. The ResumableFunctions stack frames point at the generated `MBQCPurificationTracker_FSMI`, but the source location is the `@resumable` method in `src/ProtocolZoo/mbqc.jl`, especially lines 468-471:

```julia
msg_data = querydelete!(mb, PurifierBellMeasurementResults, ❓, ❓, ❓)
local_measurements_XX = local_tag.tag[3]
local_measurements_ZZ = local_tag.tag[4]
_, (_, remote_node, remote_measurements_XX, remote_measurements_ZZ) = msg_data
```

The JET report is `convert(::Type{Tag}, ::Nothing)` / `Tag(data::Nothing)`. I think that is caused by the combination of `querydelete!` returning `Union{Nothing, NamedTuple{(:src, :tag), ...}}` and the nested destructuring of `msg_data.tag::Tag`. The preceding `query(mb, ...)` does not narrow the result of the later `querydelete!` call, and the delete query is broader than the query that found the remote message. After `@resumable` lowers this, the destructuring appears to create a `Tag`-typed temporary/iterator that is initialized with `nothing`, so JET reports the macro-generated function rather than this line.

Suggested fix: make the delete query match the checked message and avoid destructuring the `Tag` iterator, e.g.

```julia
msg_data = querydelete!(mb, PurifierBellMeasurementResults, remote_chief_idx, ❓, ❓)
isnothing(msg_data) && continue  # or @assert !isnothing(msg_data)
remote_measurements_XX = msg_data.tag[3]
remote_measurements_ZZ = msg_data.tag[4]
```

That should remove the `Tag(::Nothing)` path JET sees, and it is also more precise than deleting any `PurifierBellMeasurementResults` message from the buffer.

Krastanov (approved):

PR #286, opened 2025-10-07
Add T1T2 noise

hanakl:
The first commit was created entirely by Claude Code. It seems like it was not capable of coming up with the Kraus operators, so I added the correct operators and edited the tests.


src/backends/quantumoptics/noninstant.jl
```diff
@@ -46,3 +59,23 @@ end
...
 end
+
+"""
+Lindblad operators for combined T₁ and T₂ noise.
+
+Returns a tuple of Lindblad operators:
+- `L₁ = (1/√T₁) |0⟩⟨1|` for amplitude damping
+- `L₂ = (1/√(2Tᵩ)) Z` for pure dephasing (if T₂ < 2T₁)
+
+where `1/Tᵩ = 1/T₂ - 1/(2T₁)`
```
Krastanov:
```suggestion
Lindblad operators for combined T₁ and T₂ noise.

Returns a tuple of Lindblad operators:
- `L₁ = (1/√T₁) |0⟩⟨1|` for amplitude damping
- `L₂ = (1/√(2Tᵩ)) Z` for pure dephasing (if T₂ < 2T₁)

where `1/Tᵩ = 1/T₂ - 1/(2T₁)`.

Of note, this is **not** the same as having "on top of each other"
T₁ noise and then an additional "dephasing" noise. As you can see from
the formula above, T₁ is causing dephasing of its own.
Consult the distinction between T₂ and T₂* and Tᵩ.
```

src/backends/quantumoptics/noninstant.jl
```diff
@@ -46,3 +59,23 @@ end
...
+"""
+function lindbladop(T1T2::T1T2Noise)
+    Tᵩ_inv = 1/T1T2.t2 - 1/(2*T1T2.t1)
+
+    if Tᵩ_inv <= 0
+        return (1/√T1T2.t1 * _lh,)
+    end
```
Krastanov: could you get a reference for this choice

src/backends/quantumoptics/noninstant.jl
Krastanov: could you fill out what T2 T2* and Tphi are in the docstring, so that we can make sure I did not get it wrong

Krastanov:
@hanakl bump on this, just so it does not get lost

PR #359, opened 2026-03-20
Add DepolarizedBellPair

hanakl:
Many examples currently implement `noisy_pair_func` helper function from scratch, and there is a TODO to introduce a dedicated depolarization helper. 
This PR adds a `DepolarizedBellPair` state to the StatesZoo, representing the depolarized Bell state p ∣00⟩⟨11∣ + (1−p) I/4. The implementation supports both backends and includes accompanying tests.

Closes issue #346


Krastanov (changes requested):
- [x] switch back to testsets and rename the tests file to match the new test runner
- [x] remove old code that is not in use anymore

examples/congestionchain/setup.jl
```diff
@@ -66,7 +67,8 @@ end
 const perfect_pair = (Z1⊗Z1 + Z2⊗Z2) / sqrt(2)
 const perfect_pair_dm = SProjector(perfect_pair)
 const mixed_dm = MixedState(perfect_pair_dm)
-noisy_pair_func(F) = F*perfect_pair_dm + (1-F)*mixed_dm # TODO make a depolarization helper
+# noisy_pair_func(F) = F*perfect_pair_dm + (1-F)*mixed_dm # TODO make a depolarization helper
```
Krastanov: changing the commented out section to say something like "and here is how you can do it manually if you want to have some more general state provided by QuantumSymbolics. Check out also the StatesZoo as a source of other predefined types of noisy Bell pairs"

Krastanov (changes requested):
Looks good, feel free to merge after all tests pass (JET might be the only one failing right now) and comments below are addressed

Krastanov (approved):

PR #377, opened 2026-04-06
saving EntanglementConsumer logs

JoseRodriguezRomero:
Support for saving and loading log files of an ``EntanglementConsumer`` protocol instance. The logs can be saved in ``.txt``, ``.csv`` and ``.h5`` formats. The saved data will consist of the two observables recorded by the EntanglementConsumer and the time of each event. Moreover, for ``.h5`` formatted files, an optional parameter has been added to 
```julia
@resumable function (prot::EntanglementConsumer)(metadata::Dict{String,Any} = Dict{String,Any}())
```
to store any metadata a user may want to save (if the selected file format is ``.txt`` or ``.csv`` the metadata is not saved).


src/ProtocolZoo/ProtocolZoo.jl
```diff
@@ -27,7 +27,9 @@ export
     QDatagram, Flow, LinkLevelRequest,
     QTCPPairBegin, QTCPPairEnd,
     LinkLevelReply, LinkLevelReplyAtHop, LinkLevelReplyAtSource,
-    NetworkNodeController, EndNodeController, LinkController
+    NetworkNodeController, EndNodeController, LinkController,
+    # save and load entanglement consumer logs
+    EntanglementConsumerLog, save_entanglement_consumer_log, load_entanglement_consumer_log
```
Krastanov: let's avoid creating new public functions and structs -- they are a significant maintenance burden in the future

per-protocol log structure and saving/loading function also seems like a design that will not scale well when we add similar features to many other protocols

src/ProtocolZoo/consumer.jl
```diff
@@ -0,0 +1,274 @@
+using HDF5
+
+"""
+$TYPEDEF
+
+Structure containing all the recorded observables from an `EntanglementConsumer` protocol, along its stored metadata.
+
+$FIELDS
+"""
+@kwdef struct EntanglementConsumerLog
```
Krastanov: Introducing this new type seems to have caused the need to change a lot of code in many other unrelated parts of the library. I would like to avoid breaking changes unless they are necessary. Admittedly this is only a "technically breaking", not "formally breaking" change as it is not touching public APIs, but it is easy to avoid it, so let's avoid it.

It also happens to be a change that introduce new formally public API end points -- avoiding that when possible also helps with the review and maintenance burden, so let's do it as well.

Lastly, the previous setup conformed to the Tables.jl AbstractRow API better that makes it more easy to consume from other packages, another reason I would prefer to stick to it.

Project.toml
```diff
@@ -14,6 +14,7 @@ DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 Gabs = "0eb812ee-a11f-4f5e-b8d4-bb8a44f06f50"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 GraphsMatching = "c3af3a8c-b79e-4b01-bf44-c718d7e0e0d6"
+HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
```
Krastanov: this is a pretty significant dependency, let's keep it in a pkg extension

src/ProtocolZoo/consumer.jl
```diff
@@ -0,0 +1,274 @@
...
+function _save_log(file_name::Union{String,Nothing}, log::EntanglementConsumerLog)
+    if isnothing(file_name)
+        return
+    end
+
+    if endswith(file_name, ".h5") || endswith(file_name, ".hdf5")
+        _save_log_hdf5(file_name, log)
+    elseif endswith(file_name, ".csv")
+        _save_log_csv(file_name, log)
+    elseif endswith(file_name, ".txt")
```
Krastanov: txt is not a standard format, let's just remove it and keep only well defined standard formats

src/ProtocolZoo/consumer.jl
```diff
@@ -0,0 +1,274 @@
...
+        end
+
+        write(file, "metadata_keys", metadata_keys)
+        write(file, "metadata_values", metadata_values)
+    end
+end
+
+function _save_log_csv(file_name::String, log::EntanglementConsumerLog)
+    open(file_name, "w") do file
+        println(file, "time,obs1,obs2")
```
Krastanov: use CSV.jl, do not hand-roll your own solution (put it in a pkg extension)

src/ProtocolZoo/consumer.jl
```diff
@@ -0,0 +1,274 @@
...
+        pretty_table(file, hcat(log.time, log.obs1, log.obs2), header=["time", "obs1", "obs2"], formatters=ft_printf("%16.8f"))
+    end
+end
+
+function _save_log(file_name::Union{String,Nothing}, log::EntanglementConsumerLog)
+    if isnothing(file_name)
+        return
+    end
+
+    if endswith(file_name, ".h5") || endswith(file_name, ".hdf5")
```
Krastanov: this type of switch statements (if/else sequences) are usually a sign that multiple dispatch should be used. Check the FileIO for a more central API for these that will be more scallable and extendable, with fewer if-else trees, e.g. see the `save` block here http://juliaio.github.io/FileIO.jl/stable/implementing/#All-at-once-I/O:-implementing-load-and-save

E.g. if HDF5 and CSV in julia are both built around the FileIO ecosystem (check if this is the case), then the proper solution would be to define a

`FileIO.save(::Format{CSV}, ::EntanglementConsumer)`

Krastanov (changes requested):
Thanks for setting this up, it seems to be going in the right direction.

I have a few suggestions that should hopefully make the code simpler, shorter, and more local (not changing so many use locations). You can re-request review when they are done.

Krastanov:
Let's also remove the metadata dict until we have an explicit use for it. It seems like something that might be more appropriate in the save function itself, not in the protocol.

src/ProtocolZoo/ProtocolZoo.jl
JoseRodriguezRomero: Reverted to how it was before. As of the writing of this comment, the test scripts still need to be reverted back (I will do this tomorrow).

The main issue with the current implementation is that there is no "simple" way to tell when the simulation has finally stopped, or at least I cannot easily see how, without having access to an exported save function to store the log in a file.

src/ProtocolZoo/consumer.jl
JoseRodriguezRomero: Removed the new type.

Project.toml
JoseRodriguezRomero: added extensions for HDF5 and CVS types.

src/ProtocolZoo/consumer.jl
JoseRodriguezRomero: removed saving to .txt format.

src/ProtocolZoo/consumer.jl
JoseRodriguezRomero: added support to saving to CSV as an extension using CSV.jl

src/ProtocolZoo/consumer.jl
JoseRodriguezRomero: Added save functions as suggested.

src/ProtocolZoo/ProtocolZoo.jl
JoseRodriguezRomero: The test scripts have been reverted back to how they were before.

Krastanov:
I edited the title to not be the branch name -- the titles get auto-listed in release notes so it helps to have them be explanatory

ext/QuantumSavoryCSV/QuantumSavoryCSV.jl
```diff
@@ -0,0 +1,13 @@
+module QuantumSavoryCSV
+
+import QuantumSavory
+using QuantumSavory.ProtocolZoo
+import FileIO, CSV, DataFrames
```
Krastanov: dataframes is a big dependency, remove it if it can be removed

ext/QuantumSavoryHDF5/QuantumSavoryHDF5.jl
```diff
@@ -0,0 +1,27 @@
...
+
+import QuantumSavory
+using QuantumSavory.ProtocolZoo
+import FileIO, HDF5
+
+function FileIO.save(save_file::FileIO.File{FileIO.DataFormat{:HDF5},String}, prot::EntanglementConsumer; metadata::Union{Dict{String,Any},Nothing} = nothing)
+    HDF5.h5open(save_file.filename, "w") do file
+        write(file, "t", [log.t for log in prot._log])
+        write(file, "obs1", [log.obs1 for log in prot._log])
+        write(file, "obs2", [log.obs2 for log in prot._log])
```
Krastanov: need to figure out more standard names than obs1/obs2, so that it can be coordinated with the Sequence team

ext/QuantumSavoryCSV/QuantumSavoryCSV.jl
```diff
@@ -0,0 +1,13 @@
...
+
+function FileIO.save(save_file::FileIO.File{FileIO.DataFormat{:CSV},String}, prot::EntanglementConsumer; metadata::Union{Dict{String,Any},Nothing} = nothing)
+    if !isempty(prot._log)
+        CSV.write(save_file, DataFrames.DataFrame(prot._log, ["t", "obs1", "obs2"]))
```
Krastanov: obs1 and obs2 should be better named

ext/QuantumSavoryHDF5/QuantumSavoryHDF5.jl
Krastanov, line 1: we should be idiomatic about hdf5

- it seems it has the notion of group and dataset, let's make sure we use those as appropriate
- it seems to have the notion of an attribute (for metadata) that would be more natural than the adhoc  "metadata_keys"

Krastanov (changes requested):
some naming conventions and using hdf5 idiomatically seem to be the main outstanding issues

ext/QuantumSavoryCSV/QuantumSavoryCSV.jl
JoseRodriguezRomero: DataFrames is no longer a dependency for QuantumSavoryCSV.jl

ext/QuantumSavoryHDF5/QuantumSavoryHDF5.jl
JoseRodriguezRomero: the tuple names using _log of a EntanglementConsumer structure instance are used instead of hard coded names.

ext/QuantumSavoryCSV/QuantumSavoryCSV.jl
JoseRodriguezRomero: the tuple names using _log of a EntanglementConsumer structure instance are used instead of hard coded names.

ext/QuantumSavoryHDF5/QuantumSavoryHDF5.jl
JoseRodriguezRomero: The recorded data in ``_log`` of a ``EntanglementConsumer`` instance are now saved in separate datasets (the name given to each column in the ``@NamedTuple`` is automatically reused and used to identify each dataset).

Instead of using the custom made ``Dict{String,String}``, each entry in the ``metadata::Union{Dict{String,Any},Nothing}`` is saved as an attribute assigned to the main group of the file being saved. The values of these attributes are no longer casted to string formats, they keep their original variable type.

ext/QuantumSavoryHDF5/QuantumSavoryHDF5.jl
```diff
@@ -0,0 +1,24 @@
...
+            end
+        end
+
+        if !isempty(prot._log)
+            for property_name in propertynames(prot._log[1])
+                write(file, string(property_name), [getfield(log, property_name) for log in prot._log])
+            end
+        end
+    end
+end
```
Krastanov, line 119: we need to figure out a good standard structure for this one

Krastanov: try to follow this format:

```
Attribute: format_version :: UInt64
Attribute: format_version_minor :: UInt64 (same as semver)
Attribute: reference_state :: Enum
Committed datatype: reference_state_dt an enum type with only one entry “bellpair”
Attribute: log_format :: Enum
Committed datatype: log_format_dt an enum type with entries “fidelity”, “werner_vector”, “pauli_observables”, “state_vector”, “density_matrix” (lets worry only about pauli_observables)
Group: simulation_log
Dataset: time :: Vector{Float64}
Dataset: state :: type depends on the log_format enum
If it is pauli_observables, the index format should be 2D array of Float64
Dense with respect to the index of the observable
Strided with respect to the index of the sample
Group: metadata
Attribute: description :: str
Attribute: simulator :: str
Group: quantumsavory
Attribute: version

```

JoseRodriguezRomero: The HDF5 files are now saved with the structure we coordinated with the SeQUeNCe group

test/general/aqua_tests.jl
```diff
@@ -3,10 +3,21 @@ using Aqua
 using QuantumOpticsBase
 using QuantumSavory
 using Gabs
+using WorkerUtilities
 
 @testset "Aqua" begin
 
-@test Test.detect_ambiguities(QuantumSavory) == Tuple{Method, Method}[]
```
Krastanov, line 9: An exclude would be much simpler (single line chance) and would not require adding WorkerUtilities to the Project.toml. See the documentation of detect_ambiguities

https://juliatesting.github.io/Aqua.jl/dev/ambiguities/#Aqua.test_ambiguities

JoseRodriguezRomero: I initially had this script in mind

```julia
using Test
using Aqua
using QuantumOpticsBase
using QuantumSavory
using Gabs

@testset "Aqua" begin

@test isempty(Test.detect_ambiguities(QuantumSavory; exclude=[WorkerUtilities.put!]))

Aqua.test_all(QuantumSavory,
    ambiguities=(QuantumSavory; recursive=false),
    piracies=(; treat_as_own=[QuantumSavory.Symbolic, QuantumOpticsBase.Ket, QuantumOpticsBase.Operator, Gabs.GaussianChannel, Gabs.GaussianState, Gabs.GaussianUnitary]),
    stale_deps=(; ignore=[:NetworkLayout]) # needed by package extension but not a condition of its loading
)

@test length(Aqua.Piracy.hunt(QuantumSavory)) == 8 # TODO upstream the sources of piracies

end
```

but when I run it, it crashes with an error, saying that `WorkerUtilities` is not defined. 
```julia
Aqua: Error During Test at /Users/j.romero/Documents/GitHub/QuantumSavory.jl/test/general/aqua_tests.jl:21
  Test threw exception
  Expression: isempty(Test.detect_ambiguities(QuantumSavory; exclude = [WorkerUtilities.put!]))
  UndefVarError: `WorkerUtilities` not defined in `Main`
  Suggestion: check for spelling errors or missing imports.
  Stacktrace:
   [1] top-level scope
     @ ~/Documents/GitHub/QuantumSavory.jl/test/general/aqua_tests.jl:21
   [2] macro expansion
     @ ~/.julia/juliaup/julia-1.12.6+0.aarch64.apple.darwin14/share/julia/stdlib/v1.12/Test/src/Test.jl:1777 [inlined]
   [3] macro expansion
     @ ~/Documents/GitHub/QuantumSavory.jl/test/general/aqua_tests.jl:21 [inlined]
   [4] macro expansion
     @ ~/.julia/juliaup/julia-1.12.6+0.aarch64.apple.darwin14/share/julia/stdlib/v1.12/Test/src/Test.jl:677 [inlined]
```

JoseRodriguezRomero: I think that the ideal scenario would be to have #424 fixed before this PR, but right now I don't know if this would be a small fix thing to fix.

JoseRodriguezRomero: changing from `Test.detect_ambiguities(QuantumSavory) == Tuple{Method, Method}[]` to  `Aqua.test_ambiguities(QuantumSavory)` removes all errors in my local machine.

Krastanov: To your first comment, `put!` is **not** in WorkerUtilities.

```
julia> @which put!
Base
```

But if you have found some other way to clean up the tests, then great!

test/general/aqua_tests.jl
```diff
@@ -6,7 +6,7 @@ using Gabs
...
+Aqua.test_ambiguities(QuantumSavory)
 
 Aqua.test_all(QuantumSavory,
     ambiguities=(QuantumSavory; recursive=false),
```
Krastanov, line 12: you probably can simplify this further just by removing the special case for test_ambiguities now that it is working fine with default args

test/general/aqua_tests.jl
JoseRodriguezRomero: Yeah I did notice that mistake later, although changing that line to
```julia
@test isempty(Test.detect_ambiguities(QuantumSavory; exclude=[Base.put!]))
```
still won't work since `exclude` is not a valid argument for `Test.detect_ambiguities`, this is only supported in `Aqua.test_ambiguities`.

test/general/aqua_tests.jl
Krastanov, line 12:
```suggestion
Aqua.test_all(QuantumSavory,
```

test/general/entanglementconsumer_log_save_and_load_tests.jl
Krastanov, line 1: clean this up, it has a bunch of commented out code that makes it unnecessarily difficult to read

test/general/entanglementconsumer_log_save_and_load_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            "T2" => T2,
+            "F" => F,
+            "entangler_wait_time" => entangler_wait_time,
+            "entangler_busy_time" => entangler_busy_time,
+            
+            # Mandatory metadata
+            "description" => "Simulation of a quantum network with $(nv(network)) nodes and $(ne(network)) edges, using the specified parameters.",
+            "simulator" => "QuantumSavory.jl",
+            "reference_state" => "bell_pair",
+            "log_format" => "pauli_observables"
```
Krastanov, line 87: all of this should be automatic, not something that the user can set

test/general/entanglementconsumer_log_save_and_load_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+        metadata = Dict(
+            # Optional metadata
+            "sizes" => sizes,
+            "T2" => T2,
+            "F" => F,
+            "entangler_wait_time" => entangler_wait_time,
+            "entangler_busy_time" => entangler_busy_time,
+            
+            # Mandatory metadata
+            "description" => "Simulation of a quantum network with $(nv(network)) nodes and $(ne(network)) edges, using the specified parameters.",
```
Krastanov, line 84: default to it being empty if not provided

test/general/entanglementconsumer_log_save_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            "F" => F,
+            "entangler_wait_time" => entangler_wait_time,
+            "entangler_busy_time" => entangler_busy_time,
+            
+            # Mandatory metadata
+            "description" => "Simulation of a quantum network with $(nv(network)) nodes and $(ne(network)) edges, using the specified parameters.",
+            "simulator" => "QuantumSavory.jl",
+            "reference_state" => "bell_pair",
+            "log_format" => "pauli_observables"
+        )
```
Krastanov: it should be keyed by a symbol, not a string

the API should work no matter whether we provide a dictionary or a named tuple or another iterator over symbol-value pairs

test/general/entanglementconsumer_log_save_and_load_tests.jl
JoseRodriguezRomero: default values for mandatory metadata fields are now given in `QuantumSavoryFileIO.jl`.

test/general/entanglementconsumer_log_save_and_load_tests.jl
JoseRodriguezRomero: default values for mandatory metadata fields are now given in `QuantumSavoryFileIO.jl.jl`.

test/general/entanglementconsumer_log_save_and_load_tests.jl
JoseRodriguezRomero: comments have been removed.

test/general/entanglementconsumer_log_save_tests.jl
JoseRodriguezRomero: metadata can now be provided either as a `Dict{String,Any}` or with a `NamedTuple`.

PR #383, opened 2026-04-29
Fix lost wake-up race in `SimpleAsymmetricSemaphore`

giovanniligato:
## Motivation and workflow

I ran into this bug while implementing a distributed entanglement-distribution protocol on top of QuantumSavory. To stress-test the protocol I was running it on randomly generated configurations: most of them terminated correctly, but one specific configuration consistently deadlocked the simulation.

Because the failure was a silent hang rather than an exception, it was not obvious whether the problem lived in my protocol or in the framework underneath. I asked Claude Opus to help me investigate. After enabling verbose debug output across my protocol and going through the trace together, we narrowed the symptom down to the interaction between the *tagging* path (`tag!`) and the *waiting* path (`onchange` / `lock`): a process was waiting on a register, another process tagged that same register, and yet the waiter was never resumed. A wake-up was clearly being dropped somewhere below my code.

Claude Opus then pointed at `SimpleAsymmetricSemaphore` in `src/semaphore.jl` and proposed the one-line fix discussed below. After convincing myself that the analysis was correct, I applied the patch locally and re-ran my protocol *without changing anything else on my side*. The previously deterministic deadlock disappeared, and the many additional random configurations I tested afterwards all terminated correctly as well. No further deadlocks have been observed since.

## Bug description

`SimpleAsymmetricSemaphore` uses an `Int` field `nbwaiters` to let `unlock` decide whether there is anyone to wake up. Before this patch the code was:

```julia
function Base.lock(s::SimpleAsymmetricSemaphore)
    return @process _lock(s.lock.env, s)
end

@resumable function _lock(sim, s::SimpleAsymmetricSemaphore)
    s.nbwaiters += 1            # incremented INSIDE the spawned process
    @yield lock(s.lock)
    s.nbwaiters -= 1
    if s.nbwaiters > 0
        unlock(s.lock)
    else
        s.unlocking = false
    end
end

function unlock(s::SimpleAsymmetricSemaphore)
    if s.nbwaiters > 0
        s.unlocking = true
        unlock(s.lock)
    end
end
```

`Base.lock` returns a `Process` that the caller is expected to `@yield` on, but `nbwaiters` is incremented only *inside* the body of the `@resumable` `_lock`, i.e. **after the Process has been scheduled and eventually picked up by the ConcurrentSim scheduler**. There is therefore a window during which the calling process has formally requested the lock — and any future notification is meant for it — while `nbwaiters` is still `0`.

If another process calls `unlock(s)` during that window, the early-return guard `if s.nbwaiters > 0` is false and the unlock is silently dropped. The notification is lost forever.

The full deterministic interleaving observed in my simulation is:

1. Process A executes `@yield onchange(reg)` → `lock(reg.tag_waiter[])` → `lock(simple_sem)` → `@process _lock(...)`. A is now suspended on the returned Process, but `nbwaiters` is still `0`.
2. Process B calls `tag!(slot, …)` at the same simulation timestamp, which internally invokes `unlock(reg.tag_waiter[])` → `unlock(simple_sem)`. The guard sees `nbwaiters == 0`, so the unlock is a no-op even though A is genuinely waiting.
3. The scheduler eventually runs the `_lock` body: it does `nbwaiters += 1` and then `@yield lock(s.lock)` — but `s.lock` is held and was never released, so `_lock` blocks indefinitely.
4. In configurations where the only producer of the awaited event is the `tag!` from step 2, no later `unlock` arrives. The simulation hangs.

Because `AsymmetricSemaphore` is just a pair of `SimpleAsymmetricSemaphore`s that forwards `lock` / `unlock`, the race propagates to every API built on top of it — most notably `onchange(::Register)` / `onchange(::RegRef)`, used as the wait condition in the canonical `while isnothing(result); @yield onchange(reg); …; end` query loop recommended throughout the QuantumSavory and ProtocolZoo docstrings.

## No reproducer

Triggering the bug requires the specific ConcurrentSim interleaving described above, and small toy scripts almost always schedule the `tag!`-side `unlock` and the `onchange`-side `_lock` in an order that avoids the race. The hang only shows up reliably inside larger multi-process simulations with many events at the same timestamp — the pattern typical of distributed protocols. No minimal synthetic reproducer is therefore provided.

The external symptom is clear: `@yield onchange(reg)` never returns even though a matching `tag!` has already been placed on the register.

## The fix

Move the `nbwaiters` increment out of the `@resumable` `_lock` body and into the synchronous entry point `Base.lock`, so that `nbwaiters` is observably non-zero by the time the caller's `@yield` suspends:

```julia
function Base.lock(s::SimpleAsymmetricSemaphore)
    s.nbwaiters += 1                    # incremented synchronously
    return @process _lock(s.lock.env, s)
end

@resumable function _lock(sim, s::SimpleAsymmetricSemaphore)
    @yield lock(s.lock)                 # nbwaiters already accounted for
    s.nbwaiters -= 1
    if s.nbwaiters > 0
        unlock(s.lock)
    else
        s.unlocking = false
    end
end
```

Why this closes the race: `Base.lock` is called synchronously from the caller's `@resumable` body, *before* the caller yields. Once it returns, `nbwaiters` has already been bumped. Any subsequent `unlock(s)` — even one issued at the same simulation timestamp, before `_lock` has had a chance to run — therefore observes `nbwaiters > 0`, takes the active branch, and forwards the signal to the underlying `Resource`. The matching decrement stays inside `_lock`, right after `@yield lock(s.lock)` returns, so the counter still reflects the number of pending waiters at every observable point.

Because `AsymmetricSemaphore` is implemented as a pair of `SimpleAsymmetricSemaphore`s, fixing the simple variant is sufficient: every API that uses it — notably `onchange(::Register)` / `onchange(::RegRef)` via `register.tag_waiter[]`, and any direct user of `AsymmetricSemaphore` — inherits the fix automatically.

## No regression test

For the same reason no minimal reproducer exists, a deterministic regression test is impractical: any such test would rely on a specific ConcurrentSim interleaving that is not guaranteed by the public API and could silently pass on future releases. The fix has instead been validated empirically against the protocol where the bug was first observed, where the deadlock disappeared and many additional random configurations kept terminating correctly after the patch was applied.

## Acknowledgements

The root-cause analysis and the proposed one-line fix were developed together with Claude Opus, which I used as a debugging partner once it became clear that the deadlock was not in my protocol code. This formalization was also produced in collaboration with Claude Opus.

---

- [ ] Substantial new functionality is documented within the docs.
- [ ] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them. <small>There will be plenty of old code that is flagged as we are slowly transitioning to enforced formatting. Please do not worry about or address older formatting issues -- keep your PR just focused on your planned contribution.</small>

### DIFF
--- a/src/semaphore.jl
+++ b/src/semaphore.jl
@@ -10,11 +10,11 @@ end
 SimpleAsymmetricSemaphore(sim) = SimpleAsymmetricSemaphore(0, false, Resource(sim,1,level=1)) # start locked
 
 function Base.lock(s::SimpleAsymmetricSemaphore)
+    s.nbwaiters += 1
     return @process _lock(s.lock.env, s)
 end
 
 @resumable function _lock(sim, s::SimpleAsymmetricSemaphore)
-    s.nbwaiters += 1
     @yield lock(s.lock)
     s.nbwaiters -= 1
     if s.nbwaiters > 0
--- a/src/semaphore.jl
+++ b/src/semaphore.jl
@@ -5,9 +5,10 @@ because the semaphore will continue unlocking processes until all waiting proces
 mutable struct SimpleAsymmetricSemaphore # An equivalent, allocating, simpler implementation of this capability is in wait(::MessageBuffer)
     nbwaiters::Int
     unlocking::Bool
+    parent::Any                       # back-pointer to the owning AsymmetricSemaphore (or nothing)
     const lock::Resource
 end
-SimpleAsymmetricSemaphore(sim) = SimpleAsymmetricSemaphore(0, false, Resource(sim,1,level=1)) # start locked
+SimpleAsymmetricSemaphore(sim) = SimpleAsymmetricSemaphore(0, false, nothing, Resource(sim,1,level=1)) # start locked
 
 function Base.lock(s::SimpleAsymmetricSemaphore)
     s.nbwaiters += 1
@@ -21,6 +22,14 @@ end
         unlock(s.lock)
     else
         s.unlocking = false
+        # Cascade ended. If the owning AsymmetricSemaphore had unlock signals
+        # dropped while this side was cascading, fire one of them now so the
+        # waiters that re-locked into the other sub-semaphore (and any new
+        # waiters that arrived since) actually get woken.
+        p = s.parent
+        if p !== nothing
+            _drain_pending_unlock(p::AsymmetricSemaphore)
+        end
     end
 end
 
@@ -41,12 +50,24 @@ end
 
 Internally, it is implemented as a pair of SimpleAsymmetricSemaphores -- whenever one of them is unlocked,
 we switch to the other one, thus avoiding infinite loops when a process waits on the semaphore and then immediately starts waiting on it again.
+
+Unlock calls that arrive while a cascade is in progress are queued in
+`pending_unlocks` and replayed when the cascade ends, so wakeup signals are
+never silently dropped.
 """
 mutable struct AsymmetricSemaphore # An equivalent, allocating, simpler implementation of this capability is in wait(::MessageBuffer)
     current_semaphore::Int
+    pending_unlocks::Int
     const semaphorepair::Tuple{SimpleAsymmetricSemaphore, SimpleAsymmetricSemaphore}
 end
-AsymmetricSemaphore(sim) = AsymmetricSemaphore(1, (SimpleAsymmetricSemaphore(sim), SimpleAsymmetricSemaphore(sim)))
+function AsymmetricSemaphore(sim)
+    s1 = SimpleAsymmetricSemaphore(sim)
+    s2 = SimpleAsymmetricSemaphore(sim)
+    parent = AsymmetricSemaphore(1, 0, (s1, s2))
+    s1.parent = parent
+    s2.parent = parent
+    return parent
+end
 
 function Base.lock(s::AsymmetricSemaphore)
     sem = s.semaphorepair[s.current_semaphore]
@@ -54,7 +75,30 @@ function Base.lock(s::AsymmetricSemaphore)
 end
 
 function unlock(s::AsymmetricSemaphore)
-    if !(s.semaphorepair[1].unlocking || s.semaphorepair[2].unlocking)
+    if s.semaphorepair[1].unlocking || s.semaphorepair[2].unlocking
+        # A cascade is in progress on at least one sub-semaphore. Triggering a
+        # second cascade now would either be silently merged into the running
+        # one (if it lands on the same sub-sem) or create concurrent cascades
+        # that re-feed each other (if it lands on the other side, since the
+        # newly toggled `current` would be the still-cascading one). Queue the
+        # unlock instead and replay it from `_lock` when the running cascade
+        # completes.
+        s.pending_unlocks += 1
+        return nothing
+    end
+    sem = s.semaphorepair[s.current_semaphore]
+    s.current_semaphore = 3 - s.current_semaphore
+    unlock(sem)
+    return nothing
+end
+
+# Called by `_lock` when its sub-semaphore's cascade ends. Fires one queued
+# unlock against the (now fresh) opposite sub-semaphore. Each queued unlock
+# eventually surfaces because every cascade ends in `_lock`, which calls back
+# into this drain.
+function _drain_pending_unlock(s::AsymmetricSemaphore)
+    if s.pending_unlocks > 0 && !(s.semaphorepair[1].unlocking || s.semaphorepair[2].unlocking)
+        s.pending_unlocks -= 1
         sem = s.semaphorepair[s.current_semaphore]
         s.current_semaphore = 3 - s.current_semaphore
         unlock(sem)


Krastanov:
Oh, this looks incredibly helpful, thank you so much for putting the time to investigate it! The AsymmetricSemaphore has been generally a pretty tough piece of code to implement and properly take care of all edge cases.

Do you have a minimal example that we can put in the tests so that we can avoid breaking it in the future?

Krastanov:
On a separate note, you have had a few very helpful remarks/reports/contributions. The team is certainly interested in engaging with folks like that (especially given that most of this is volunteer run). If you are interested in becoming member of the github organization (to help with the occasional PR review) or if you would like to engage in some paid summer project related to this software, let me know. None of this is really a formal commitment -- plenty of folks join open source projects, help with a couple of PR reviews or issues on and off over a few months or years, and then disappear, which is perfectly fine.

giovanniligato:
> Oh, this looks incredibly helpful, thank you so much for putting the time to investigate it! The AsymmetricSemaphore has been generally a pretty tough piece of code to implement and properly take care of all edge cases.
> 
> Do you have a minimal example that we can put in the tests so that we can avoid breaking it in the future?

Thanks for your response! I did try to put together a minimal reproducer before opening the PR, precisely because I know how important it is for showing the bug and locking in a regression test, but I quickly encountered the problem mentioned in the PR: with only a handful of entities the scheduler tends to order things in the straightforward way and the deadlock simply does not occur.

What I can do is take the specific scenario from my distribution protocol where the bug shows up and strip it down as much as possible while keeping the deadlock visible. It won't be truly minimal, several entities will need to remain, and I cannot fully guarantee its determinism across future releases of the components involved, since it relies on the current scheduler picking a specific interleaving; a different scheduler implementation could happily reorder things and hide the deadlock again even if the underlying bug were reintroduced.

Nevertheless I'm Happy to walk that path if you think it's worth it, or to try a different approach if you have one in mind, just let me know!

> On a separate note, you have had a few very helpful remarks/reports/contributions. The team is certainly interested in engaging with folks like that (especially given that most of this is volunteer run). If you are interested in becoming member of the github organization (to help with the occasional PR review) or if you would like to engage in some paid summer project related to this software, let me know. None of this is really a formal commitment -- plenty of folks join open source projects, help with a couple of PR reviews or issues on and off over a few months or years, and then disappear, which is perfectly fine.

It's really an honor to receive a message like this, thank you.

Would it be possible to consider this opportunity again in the coming months? I'm currently in the final stage of my Master's thesis here at the University of Pisa, and as I get closer to the deadline my workload is increasing quite a bit. I'd prefer to think about it more carefully once things settle, possibly after the summer.

Thanks again for the opportunity. In the meantime, I'll continue working with QuantumSavory as I complete my protocol, so I can further deepen my understanding of the framework.

giovanniligato:
@Krastanov sorry to bother you again, but I wanted to share an update. Before opening this PR I pulled my local QuantumSavory dev checkout to the latest master and re-ran my protocol with the fix proposed here applied on top. The race this PR addresses is indeed gone, but today I ran into a second deadlock that was not present in the older checkout I had been using (about a month behind master).

I narrowed the regression down to commit `a88d7a2116cc9eb0b1bf717b86cfc08eee391c41` ("use ASemaphore in messagebuffer as well"). Reverting to immediately before that commit makes the deadlock disappear; reapplying it brings it back deterministically on the same configuration.

The symptom is a hang in a `MessageBuffer` waited on by several processes at the same node: when multiple `put!`s land in the same simulation timestamp during an in-flight cascade, the corresponding `unlock(::AsymmetricSemaphore)` calls hit the `if !(s1.unlocking || s2.unlocking)` guard and return silently, dropping the wakeup. The waiters that re-locked into the now-current sub-semaphore during the cascade are then never resumed and their messages rot in the buffer. The `nbwaiters` fix in this PR is necessary but does not cover this case, since the dropped wakeup is at the pair layer, not at the simple-semaphore layer.

I have already prototyped a fix and verified it against my protocol. QuantumSavory's existing semaphore tests still pass.

Since the fix is structurally separate from the `nbwaiters` one (it touches the `AsymmetricSemaphore` pair, not `SimpleAsymmetricSemaphore`) would you prefer I add it as an extra commit on this PR, or open a separate PR?

Krastanov:
Thank you so much for looking into this as well, this help is absolutely invaluable! Please commit that fix here as well, that way we can brainstorm appropriate tests for both related fixes and potentially some simplifications as well.

giovanniligato:
# Fix dropped wake-ups in `AsymmetricSemaphore` during a cascade

## Motivation and workflow

After pulling the latest changes on top of `nbwaiters` fix to `SimpleAsymmetricSemaphore`, I observed a new deadlock in the same distributed entanglement-distribution simulation that motivated that earlier fix. The earlier fix is still necessary and still resolves the race it was written for: this report is **not** about that bug resurfacing. The new failure mode appeared specifically after pulling commit `a88d7a2116cc9eb0b1bf717b86cfc08eee391c41` (2026-03-31).  Before that commit my simulation ran cleanly with the `nbwaiters` fix applied; after pulling it, a specific concurrent interleaving (many processes blocked on the same `MessageBuffer`, multiple messages arriving at the same simulation timestamp) consistently hangs after the first hand-off, with the entry-point process never returning a result.

I went through the trace with Claude Opus. Adding verbose logging to my protocol's wait-loops we observed that the `MessageBuffer` at one node held a message that no waiter ever consumed — even though several waiters were blocked on the corresponding `tag_waiter`. After ruling out the protocol code (the wait pattern is the canonical `while isnothing(r); @yield onchange(mb); r = querydelete!(mb, …); end` documented across the codebase), the root cause was located in `AsymmetricSemaphore.unlock`. We applied the patch described below and the deadlock disappeared across many random executions of the simulation that previously failed; QuantumSavory's own `semaphore_tests.jl` (68 assertions, including the multi-waiter / re-yield torture tests) keeps passing.

## Bug description

`AsymmetricSemaphore` is a pair of `SimpleAsymmetricSemaphore`s with a `current_semaphore` index. Calls to `Base.lock` register on the *current* sub-semaphore; calls to `unlock` swap which sub-semaphore is current and unlock the previous one. The pair design exists to avoid the documented infinite-loop hazard of `SimpleAsymmetricSemaphore`: when a waiter wakes up and *immediately* calls `lock` again (the canonical `MessageBuffer` re-query pattern), it must land on a *different* sub-semaphore so the ongoing cascade does not re-wake it.

After commit `a88d7a21…` the unlock guard is:

```julia
function unlock(s::AsymmetricSemaphore)
    if !(s.semaphorepair[1].unlocking || s.semaphorepair[2].unlocking)
        sem = s.semaphorepair[s.current_semaphore]
        s.current_semaphore = 3 - s.current_semaphore
        unlock(sem)
    end
    return nothing
end
```

If *either* sub-semaphore is already mid-cascade (`semaphorepair[i].unlocking == true`), the call is silently dropped. The intent is to prevent two concurrent cascades from feeding each other (a re-locker landing on a still-cascading sub-sem perpetuates that cascade). The flaw is that the dropped call carried a real wake-up signal that nobody else will replay. From the caller's perspective, `unlock` returns successfully whether it actually scheduled a wake-up or threw it away, so any code that relies on "every `unlock` produces at least one observable wake-up" is silently miscompiled into "most `unlock`s produce a wake-up".

The deterministic interleaving observed in my simulation is:

1. Waiters W₁, …, Wₖ are blocked on sub-semaphore 1 (`semaphorepair[1].nbwaiters == k`, `current_semaphore == 1`).
2. Process A calls `put!(mb, tag₁)` → `unlock(mb.tag_waiter::AsymmetricSemaphore)`. Neither sub-sem is cascading, so the unlock fires: `current_semaphore` becomes 2, `semaphorepair[1].unlocking` becomes `true`, the underlying `Resource` is released, and the cascade through sub-sem 1's waiters begins.
3. As the cascade unfolds, the woken Wᵢ that do not match tag₁ re-enter the `while isnothing(r) … @yield onchange(mb) … end` loop. Each such re-entry calls `Base.lock(::AsymmetricSemaphore)`, which registers on the *current* sub-semaphore — i.e. sub-sem 2 — incrementing `semaphorepair[2].nbwaiters`.
4. While the sub-sem-1 cascade is still in flight, process B calls `put!(mb, tag₂)`. `put_and_unlock_waiters` sees `nbwaiters(mb.tag_waiter) > 0` (because some Wᵢ have already re-locked into sub-sem 2) and takes the `unlock(mb.tag_waiter)` branch.
5. In `AsymmetricSemaphore.unlock` the guard `s.semaphorepair[1].unlocking == true` fires, so the call returns silently. There is no replay queue; the wake-up signal for tag₂ is lost.
6. Sub-sem 1's cascade eventually completes. Sub-sem 2 has `nbwaiters > 0` and a matching message in the buffer, but no further `unlock` arrives for it (in the deadlocking configurations no later message is destined for that node, so step 4 is the last `put!`). The simulation has no more events to process and the entry-point process never returns.

The same race propagates to `Register.tag_waiter`, which is also an `AsymmetricSemaphore` driven by `tag!` / `untag!`. Any caller of the canonical `while isnothing(r); @yield onchange(reg); r = querydelete!(reg, …); end` loop is exposed under sufficient timestamp-collision pressure.

## No reproducer

As with the earlier `nbwaiters` race, triggering this requires a specific ConcurrentSim interleaving — many waiters on the same `MessageBuffer`, several `put!`s landing in the same simulation timestamp, and at least one of those `put!`s observing a sub-semaphore that is already cascading. Small synthetic scripts almost always serialize the events into an order that avoids step 4. The hang only shows up reliably inside larger multi-process simulations with many same-timestamp events — the typical shape of distributed protocols. No minimal synthetic reproducer is therefore provided.

The external symptom is identical to the earlier bug: `@yield onchange(mb)` (or `@yield onchange(reg)`) never returns even though a matching message or tag has already been placed.

## The fix

Stop dropping unlocks. When an unlock arrives during an ongoing cascade, queue it on a counter on the parent `AsymmetricSemaphore`; when the cascade ends inside `_lock`, drain one queued unlock against the (now fresh) opposite sub-semaphore. Each replayed unlock itself ends in a `_lock` callback, so a chain of queued unlocks drains without further intervention.

```julia
mutable struct SimpleAsymmetricSemaphore
    nbwaiters::Int
    unlocking::Bool
    parent::Any                       # back-pointer to the owning AsymmetricSemaphore (or nothing)
    const lock::Resource
end
SimpleAsymmetricSemaphore(sim) =
    SimpleAsymmetricSemaphore(0, false, nothing, Resource(sim, 1, level=1))

@resumable function _lock(sim, s::SimpleAsymmetricSemaphore)
    @yield lock(s.lock)
    s.nbwaiters -= 1
    if s.nbwaiters > 0
        unlock(s.lock)
    else
        s.unlocking = false
        p = s.parent
        if p !== nothing
            _drain_pending_unlock(p::AsymmetricSemaphore)
        end
    end
end

mutable struct AsymmetricSemaphore
    current_semaphore::Int
    pending_unlocks::Int
    const semaphorepair::Tuple{SimpleAsymmetricSemaphore, SimpleAsymmetricSemaphore}
end
function AsymmetricSemaphore(sim)
    s1 = SimpleAsymmetricSemaphore(sim)
    s2 = SimpleAsymmetricSemaphore(sim)
    parent = AsymmetricSemaphore(1, 0, (s1, s2))
    s1.parent = parent
    s2.parent = parent
    return parent
end

function unlock(s::AsymmetricSemaphore)
    if s.semaphorepair[1].unlocking || s.semaphorepair[2].unlocking
        s.pending_unlocks += 1
        return nothing
    end
    sem = s.semaphorepair[s.current_semaphore]
    s.current_semaphore = 3 - s.current_semaphore
    unlock(sem)
    return nothing
end

function _drain_pending_unlock(s::AsymmetricSemaphore)
    if s.pending_unlocks > 0 &&
       !(s.semaphorepair[1].unlocking || s.semaphorepair[2].unlocking)
        s.pending_unlocks -= 1
        sem = s.semaphorepair[s.current_semaphore]
        s.current_semaphore = 3 - s.current_semaphore
        unlock(sem)
    end
    return nothing
end
```

Why this preserves the original anti-infinite-loop guarantee: `unlock` never fires *while* a sub-semaphore is cascading, so a waiter that wakes up and immediately re-locks into the freshly toggled `current` sub-semaphore is not re-woken by the same cascade. The deferred replay only happens after the in-flight cascade has fully drained (`unlocking == false` in `_lock`) and is therefore guaranteed to land on a sub-semaphore that is not cascading. By the time the replay runs, any waiter that re-locked during the previous cascade is sitting in the *other* sub-semaphore (the one the toggle moved `current` to before the queued unlock) and will be woken by this replay — exactly the wake-up that was originally requested.

Why it is sufficient: in the deadlocking trace above, step 5 now appends to `pending_unlocks` instead of returning silently. When sub-sem 1's cascade ends in `_lock`, the drain decrements `pending_unlocks`, toggles `current_semaphore` back to 1, and unlocks sub-sem 2 — releasing the re-locked waiters who then observe tag₂ in the buffer. No tag is left orphaned, no protocol-level change is required.

The same fix automatically benefits every API built on top of `AsymmetricSemaphore`: `MessageBuffer.tag_waiter`, `Register.tag_waiter[]`, and any direct user.

## No regression test

For the same reasons given for the earlier `nbwaiters` race, no deterministic regression test is practical: any such test would rely on a specific ConcurrentSim interleaving that is not guaranteed by the public API and could silently pass on a future scheduler change. The fix has instead been validated empirically:

- QuantumSavory's existing `test/general/semaphore_tests.jl` (the `AsymmetricSemaphore functionality` testset, 68 assertions, including the multi-waiter and re-yield torture tests) keeps passing.
- The deterministic deadlock that motivated the patch disappeared.
- Several random executions of the simulation that previously hung now all complete successfully with no further deadlocks observed.

## Acknowledgements

The root-cause analysis and the proposed fix were developed together with Claude Opus, used as a debugging partner once it became clear the new deadlock was again below the protocol layer. This formalization was also produced in collaboration with Claude Opus.

Krastanov:
Thank you for this! I prepared some tests in #384. Please check them out if you have the time. Your commits were moved to that PR and merged. Thank you for the help! And no worries, take your time in deciding whether you want to join the project in any capacity. The bounty program is a particularly convenient way to get rewarded for various contributions (including suggestions of your own).

Issue #131, opened 2024-07-09
More thorough benchmarks [$200]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

# Project: "More thorough benchmarks" [$200]

We have a small benchmark suite already implemented, which is executed as part of our CI runs. It is defined in the `benchmark` folder and reported for each pull request. **We would like to expand this benchmark suite** to include many more facets of this library. E.g. basic register operations using a variety of backends; queries and tags and locks on registers and channels; time to import; time to run examples. The new benchmarks should be legible, easy to follow, and organized by topic. Most of them should be microbenchmarks testing only one concept, but a few holistic benchmarks would make sense as well.

**Required skills**: Generic Julia skills.

**Reviewer**: Stefan Krastanov

**Duration**: 1 month

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

coderight1:
Hi @Krastanov , I have recently started learning Julia , and willing to work on this.

Krastanov:
Please feel free to go ahead with it. Before I reserve it for you though, let's discuss what you plan to add in this comment thread. While this is a good issue for a first-time developer, it is not an easy issue for someone who has not written any user code with this library. I would suggest trying to run a few examples and creating a simulation of something interesting first. The benchmarks are supposed to have a very thorough coverage.

coderight1:
Hi @Krastanov,

Thank you for the opportunity. I appreciate your advice and would like to proceed with the following plan for expanding the benchmark suite:

Basic Register Operations:
[...]

PR #398, opened 2026-05-12
Add register microbenchmarks

ProtonDev-sys:
## Summary
- add focused register benchmarks for creation, network construction, initialization, and CNOT application
- cover small and larger registers, mixed Clifford/QuantumMC representations, and background-noise cases
- keep the existing `creation_and_initialization/from_tests` benchmark intact for compatibility

This is a focused benchmark expansion toward #131, not a claim that the whole benchmark coverage task is complete.

## Verification
- `julia --project=benchmark -e 'using BenchmarkTools; include("benchmark/benchmarks.jl"); function run_leaves(group, path="register"); for key in keys(group); child = group[key]; childpath = string(path, "/", key); if child isa BenchmarkGroup; run_leaves(child, childpath); else; run(child; samples=1, evals=1); println(childpath); end; end; end; run_leaves(SUITE["register"])'`
- `julia --project=test test/runtests.jl general/register_interface_tests`

Refs #131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.6.1 - unreleased
 
 - **(fix)** Solving edge cases of deadlocks when simultaneously tagging and waiting on tags.
+- Add focused register microbenchmarks covering creation, network construction, initialization, and CNOT application.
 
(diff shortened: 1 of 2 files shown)


Krastanov:

PR #399, opened 2026-05-12
Add QuantumChannel benchmarks

ProtonDev-sys:
## Summary
- add a focused `quantumchannel` benchmark group
- cover QuantumChannel construction, `put!`, and full simulated send/receive transfer
- include plain, `T1Decay`, `T2Dephasing`, and network `qchannel` transfer paths

This is a small follow-up benchmark coverage slice toward #131, separate from #398.

## Verification
- `julia --project=benchmark -e 'using BenchmarkTools; include("benchmark/benchmarks.jl"); function run_leaves(group, path="quantumchannel"); for key in keys(group); child = group[key]; childpath = string(path, "/", key); if child isa BenchmarkGroup; run_leaves(child, childpath); else; run(child; samples=1, evals=1); println(childpath); end; end; end; run_leaves(SUITE["quantumchannel"])'`
- `julia --project=benchmark -e 'include("benchmark/benchmarks.jl"); println(keys(SUITE)); println(keys(SUITE["quantumchannel"]))'`
- `julia --project=test test/runtests.jl general/quantumchannel_tests`

Refs #131


Krastanov:

PR #400, opened 2026-05-12
Add quantum-state measurement benchmarks

ProtonDev-sys:
## Summary
- add a Clifford observable benchmark alongside the existing QuantumOptics observable benchmark
- add projective-measurement benchmarks for `project_traceout!` on Bell-pair and 5-qubit GHZ states
- cover both QuantumOptics and Clifford representations with fresh setup per mutating measurement benchmark

This is a small benchmark coverage slice toward #131, separate from #398 and #399.

## Verification
- `julia --project=benchmark -e 'using BenchmarkTools; include("benchmark/benchmarks.jl"); function run_leaves(group, path="quantumstates"); for key in keys(group); child = group[key]; childpath = string(path, "/", key); if child isa BenchmarkGroup; run_leaves(child, childpath); else; run(child; samples=1, evals=1); println(childpath); end; end; end; run_leaves(SUITE["quantumstates"])'`
- `julia --project=benchmark -e 'include("benchmark/benchmarks.jl"); println(keys(SUITE["quantumstates"])); println(keys(SUITE["quantumstates"]["project_traceout"]))'`
- `julia --project=test test/runtests.jl general/project_traceout_tests general/observable_tests`

Refs #131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v0.6.1 - unreleased
 
 - **(fix)** Solving edge cases of deadlocks when simultaneously tagging and waiting on tags.
+- Add observable and projective-measurement benchmarks for quantum states.
 
 ## v0.6.0 - 2026-05-05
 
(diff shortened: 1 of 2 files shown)


Krastanov:

Issue #410, opened 2026-05-14
QuantumSavory simulations sometimes crash with a KeyError message.


JoseRodriguezRomero:
# Bug Description

Some simulations of simple repeater chains crash with a ``KeyError: key 4873648 not found`` error message. I believe, this issue could be a query being untaged and marked for deletion without locking, and some other part of the code is asynchronously then attempting to fetch an already deleted element of the ``Dict`` where the tagged queries are stored.

# Minimal Reproducible Example

To run the scripts with ``N`` threads run:

```bash
julia --threads N RunSimulations.jl
```

## RunSimulations.jl

```julia
using Pkg
using Base.Threads

Pkg.activate(".")

include("setup.jl")

n_threads = nthreads()
sample_fidelities = 0.50:0.05:0.95

@info("Using $(n_threads) threads.")

for F in sample_fidelities
    sizes = zeros(Int,5) .+ 10 # Number of qubits in each register
    T2 = 100.0                 # T2 dephasing time of all qubits
    # F = 0.85                 # Fidelity of the raw Bell pairs
    entangler_wait_time = 0.1  # How long to wait if all qubits are busy before retrying entangling
    entangler_busy_time = 1.0  # How long it takes to establish a newly entangled pair
    swapper_wait_time = 0.1    # How long to wait if all qubits are unavailable for swapping
    swapper_busy_time = 0.15   # How long it takes to swap two qubits
    purifier_wait_time = 0.15  # How long to wait if there are no pairs to be purified
    purifier_busy_time = 0.2   # How long the purification circuit takes to execute 
    n_sample_files = 350       # How many files to save for each F value (should be <= n_threads)

    base_dir = "saved_data"
    if !isdir(base_dir)
        mkdir(base_dir)
    end

    base_dir = "saved_data/F_$(F)/"
    if !isdir(base_dir)
        mkdir(base_dir)
    end

    error_dir = base_dir * "error_logs/"
    if !isdir(error_dir)
        mkdir(error_dir)
    end

    error_log_files = Dict{Int,IO}()
    for thread_id in 1:n_threads
        error_log_files[thread_id] = open(error_dir * "error_log_thread_$(thread_id).txt", "w")
    end

    obs1_vecs = Dict{Int, Vector{Float64}}()
    obs2_vecs = Dict{Int, Vector{Float64}}()
    t_vecs = Dict{Int, Vector{Float64}}()
    for sample_id in 1:n_sample_files
[...]

PR #422, opened 2026-05-16
Report invalid EntanglementCounterpart reuse

Krastanov-agent:
Refs #410.

This replaces the previous same-slot SwapperProt change. A single slot with two `EntanglementCounterpart` tags is not a valid swappable pair; it means the pairwise `EntanglementCounterpart` invariant was already broken.

The likely dangerous boundary is not SwapperProt selection itself, but protocols that create or update counterpart tags after stale bookkeeping, deletion, cutoff, or tracker messages. This PR therefore keeps low-level `tag!` permissive, but adds protocol-level error logs when `EntanglerProt` or `EntanglementTracker` is about to add an `EntanglementCounterpart` to a slot that already has one. SwapperProt now also logs the invalid same-slot symptom and waits instead of self-locking on broken state.

After turning on those guards, the noisy failures were traced to the piecemaker switch example: the switch consumed its half of each Bell pair, but left the reciprocal `EntanglementCounterpart` tag on the client slot. Cleanup later traced out the client qubit with the old tag still present, so the next round reused a free slot that still carried stale Bell-pair metadata. The example now deletes the exact reciprocal client tag when the Bell pair is consumed.

The tests are deliberately contrived with public `tag!` calls to construct the broken invariant, then check that the relevant protocol reports it clearly. A separate focused piecemaker regression runs two one-client rounds and checks that no error-level counterpart-invariant logs are emitted and that the client slot has no leftover counterpart tag after cleanup.

Tested with:
```julia
Pkg.test(; test_args=["general/protocolzoo_entanglement_counterpart_invariant_tests"])
Pkg.test(; test_args=["general/protocolzoo_entanglement_counterpart_invariant_tests", "general/protocolzoo_swapper_stale_query_tests", "general/protocolzoo_entanglement_tracker_lock_gap_tests"])
Pkg.test(; test_args=["examples/piecemakerswitch_tag_cleanup_tests"])
Pkg.test(; test_args=["examples/piecemakerswitch_1_tests"])
Pkg.test(; test_args=["examples/firstgenrepeater_2_tests"])
Pkg.test(; test_args=["examples/firstgenrepeater_v2_2_tests"])
```

### DIFF
--- a/examples/piecemakerswitch/setup.jl
+++ b/examples/piecemakerswitch/setup.jl
@@ -185,6 +185,7 @@ The piecemaker protocol generates multipartite entanglement by:
                         @yield lock(net[1][n+1]) & lock(net[1][slot.idx])
                         res = fusion(net[1][n+1], net[1][slot.idx])
                         unlock(net[1][n+1]); unlock(net[1][slot.idx])
+                        querydelete!(net[1 + slot.idx][1], EntanglementCounterpart, 1, slot.idx)
                         tag!(net[1 + slot.idx][1], Tag(:updateX, Int(res))) # communicate change to client node
                         counter += 1
                         @debug "Fused client $(slot.idx) with piecemaker qubit"
--- a/src/ProtocolZoo/ProtocolZoo.jl
+++ b/src/ProtocolZoo/ProtocolZoo.jl
@@ -69,6 +69,16 @@ end
 Base.show(io::IO, tag::EntanglementCounterpart) = print(io, "Entangled to $(tag.remote_node).$(tag.remote_slot)")
 Tag(tag::EntanglementCounterpart) = Tag(EntanglementCounterpart, tag.remote_node, tag.remote_slot)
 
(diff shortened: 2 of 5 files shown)


Krastanov:
This does not make sense. If a single slot is marked through EntanglementCounterpart as entangled to two other slots, then an important invariant is already broken.

Krastanov-agent:
Reworked this PR to remove the same-slot filtering fix. The branch now treats same-slot selection as a broken `EntanglementCounterpart` invariant and reports it instead.

The code now logs when `EntanglerProt` or `EntanglementTracker` would add a counterpart tag to a slot that already has one, and SwapperProt logs/waits if it sees the invalid same-slot symptom instead of self-locking. Tests construct the broken invariant explicitly with public `tag!` calls so the expected behavior is clear.

LBacciottini:
Should we close this since we already implemented it in #440 ?

Krastanov:
indeed

PR #434, opened 2026-05-27
Benchmark: query index selectivity

LBacciottini:
## Summary

  This PR adds a focused benchmark subgroup for tag/query selectivity in `benchmark/benchmark_tagquery.jl`.

  The new benchmarks measure cases where a simulation may accumulate many metadata entries or classical messages, while protocol logic repeatedly searches for a small semantic subset. These cases are not well captured by the existing small-container tag/query benchmarks.

  ## What This Measures

  Define *tag-head* as the first field of a tag, and *single-slot* as a query that targets a specific slot inside a register.

  The new `SUITE["tagquery"]["index_selectivity"]` subgroup covers:

  - rare tag-head lookup in a register, e.g. finding `:distilled` metadata among many `:noisy` entries;
  - rare tag-head `queryall` in a register;
  - single-slot register lookup when many other slots carry the same tag head, e.g. `:entangled`;
  - single-slot `queryall`;
  - message-buffer lookup where the requested message tag-head is rare;
  - message-buffer lookup for a missing message tag-head, e.g. `:error_corrected`;
  - mutating message-buffer `querydelete!` for a rare message tag-head in a large backlog.

  These are intended as performance indicators for common simulation patterns: metadata-heavy registers, message-buffer backlogs, and repeated protocol queries over sparse semantic subsets.

  ## CHANGELOG Notes

  Benchmarking:
  - Add tag/query selectivity benchmarks for large register and message-buffer workloads with rare semantic matches.

  ## Notes

  This PR only adds benchmarks. It does not change tag/query behavior or implementation.

  A follow-up PR may introduce internal query/indexing improvements that are expected to improve these benchmark cases.

  ## Checklist

  - [ ] Substantial new functionality is documented within the docs.
  - [x] All new functionality is tested. (Not applicable here as there is no new functionality)
  - [ ] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them.

Disclosure: I have prepared this PR using Codex under strict supervision.


Krastanov:
Looks good! Be careful with the more general case, the first entry of a tag is not always of a symbol, it might be a datatype.

PR #471, opened 2026-06-17
Add MySwapperProt tutorial

LBacciottini:
Adds a tutorial for the `MySwapperProt` example discussed in the QuantumSavory manuscript.

This PR adds:
  - a runnable `examples/myswapper_tutorial` example showing a simple three-node repeater chain with explicit tags, message buffers, and a custom swapper protocol;
  - a new docs page explaining the example as a pedagogical walkthrough;
  - entries for the tutorial in both the Tutorials and How-To Guides docs menus;
  - an example test.


docs/make.jl
```diff
@@ -74,10 +74,12 @@ function main()
         "Cluster States in Atomic Memories" => "howto/colorcentermodularcluster/colorcentermodularcluster.md",
         "Entanglement Switch" => "howto/simpleswitch/simpleswitch.md",
         "Cluster-State Walkthrough" => "howto/cluster_state_walkthrough.md",
+        "Custom Swapper Protocol" => "tutorial/myswapperprot.md",
     ],
     "Tutorials" => [
         "tutorial.md",
         "Gate Duration" => "tutorial/noninstantgate.md",
+        "Custom Swapper Protocol" => "tutorial/myswapperprot.md",
```
Krastanov, line 81: This should probably be in only one of these, not in both.

I am trying to follow the diataxis style: https://diataxis.fr/

In that formalism a "how to" is supposed to be a complete end-to-end implementation of a project and a "tutorial" is supposed to be a specific single-purpose walkthrough. I am on the fence about which one this should be. Could you pick one as you see fit?

examples/myswapper_tutorial/my_swapper_prot.jl
```diff
@@ -0,0 +1,97 @@
+include("setup.jl")
```
Krastanov: I am not sure the distinction of setup and my_swapper_prot makes a lot of sense here. In other examples, setup exists to be shared between different examples. Here, (1) there is only one example and (2) setup actually depends on the struct defined in the file that imports setup, which is a bit of an antipattern.

examples/myswapper_tutorial/README.md
```diff
@@ -0,0 +1,13 @@
+# MySwapperProt tutorial
```
Krastanov, line 1: Could you add a link to the documentation page where more about this example can be seen? Other example READMEs have something like that.

Krastanov (changes requested):
Looks great, just some minor stylistic requests. Once addressed, feel free to merge without further review.

The test failures are known false positives in the Aqua -- persistent tasks test.

docs/make.jl
LBacciottini: Fixed. Only a tutorial now.

examples/myswapper_tutorial/my_swapper_prot.jl
LBacciottini: Agreed. Merged them together in a single script

examples/myswapper_tutorial/README.md
LBacciottini: Added a sentence linking to the tutorial page

Issue #218, opened 2025-04-24
Implementation of the ECC-based nested purification protocol by Narayanan's team


Krastanov:
https://arxiv.org/abs/2504.11670

![Image](https://github.com/user-attachments/assets/feea40ad-c344-4b91-90d0-ce7a426b6e06)

PR #522, opened 2026-08-07
Correct issue 301 examples and modeling guidance

### LINKED ISSUE #301: Projective Measurement of entangled qubit in Clifford representation
I am trying to implement a networking simulation to generate a certain stabiliser state in the Clifford formalism. As part of the simulation, I need to perform a CNOT gate with the target qubit being part of a Bell pair, as in

**apply!((network[ri,q1], network[ri,1]), QuantumSavory.QuantumClifford.sCNOT) ,** 

where network was set up in the Clifford representation, network[rj,1] is entangled with network[ri,1], and network[ri,q1] is supposed to be the control. This yields the following error:

_LoadError: An attempt was made to measure a projection observable while using Clifford representation for the qubits. However, the qubits that are being observed are entangled with other qubits. Currently this is not supported. Consider tracing out the extra qubits or using Pauli observables that do not suffer from this embedding limitation. Message us on the issue tracker if you want this functionality implemented._

I am unsure if there is an obvious fix, but since the error message explicitly suggests to mention this feature request here, I was wondering whether an implementation of entangled qubit measurements in the Clifford regime would be possible. If there is a simpler fix that I haven't considered, I would be happy to hear about it!

For context, the code I am working with the following code setup:

```
@resumable function bell_pair(sim, network, ri, rj)
    eprot = EntanglerProt(sim, network, ri, rj; pairstate=noisy_pair, chooseslotA=1, chooseslotB=1, rounds=1, success_prob=1.)
    @process eprot()
end

@resumable function apply_remote_CNOT(sim, network, ri, rj, q1, q2)
    [...]
    # Create Bell Pair between first qubits of registers i and j
    @process bell_pair(sim, network, ri, rj)
    [...]

    # Perform CNOT telegate between qubits q1 and q2
    apply!((network[ri,q1], network[ri,1]), QuantumSavory.QuantumClifford.sCNOT) 
    @yield timeout(sim, two_qubit_gate_execution_time)
    m1 = project_traceout!(network[ri,1], Z)
    @yield timeout(sim, projective_measurement_time)
    if m1 == 2
        apply!(network[rj,1], X)
        @yield timeout(sim, single_qubit_gate_execution_time)
    end
    # Teleported CNOT Gate
    [...]
end

@resumable function circuit(sim, network)
    initialize!(...)
    apply!(...)
    [...]
    **@yield @process apply_remote_CNOT(sim, network, 1, 2, 5, 2)**
    [...]
end

sim, network = simulation_setup(sizes, T2; representation = CliffordRepr)  
@process circuit(sim, network)
```

For some time, I thought the error came from the project_traceout! call, but it seems that the CNOT apply! call is the root.

Thank you!

Krastanov-agent:
Related to #301. Documentation/example-only draft pending the behavior and registered symbolic-lowering sequence described below.

## Summary

This PR corrects the example and modeling guidance exposed by the issue-301 investigation. It covers:

- the manual depolarized Bell-pair fidelity parameter;
- backend-neutral gates and Pauli observables;
- register occupancy checks;
- scheduler time versus register-local time;
- symbolic initialization factorization;
- Clifford trajectory sampling;
- T1/T2 background selection; and
- waiting for a one-round `EntanglerProt`.

The separate seven-qubit symbolic-lowering dependency change is implemented in [#524](https://github.com/QuantumSavory/QuantumSavory.jl/pull/524). This documentation PR does not close #301.

## Bell-pair fidelity correction

For `ρ = p ρBell + (1-p) I/4`, the Bell overlap is `F = (3p+1)/4`, so the correct mixture weight is:

```julia
p = (4F - 1) / 3
```

The executable low-level repeater setup and matching guides use this relation and continue to recommend `DepolarizedBellPair(; F)` as the shorter route. Example tests initialize both hand-written forms and verify through public `observable` calls that the Bell-projector overlap is the requested `F`.

## Backend-neutral and register guidance

The symbolic frontend recommends exported `H`, `CNOT`, and symbolic `X`, `Y`, `Z`, and `I` tensor factors. Native `sHadamard`, `sCNOT`, and `PauliOperator` values are described as backend-specific interoperability escape hatches.

A `RegRef` is a slot view, not an occupancy sentinel. The documented checks are `isassigned(ref)` and `isassigned(register, slot)`.

The text distinguishes the two-factor symbolic route from the three-or-more-factor, identity-containing, and native-Pauli observable paths covered by registered QuantumSymbolics 0.4.17 and [#524](https://github.com/QuantumSavory/QuantumSavory.jl/pull/524).

## Time and factorization guidance

The guides now state that:

- `ConcurrentSim.timeout` advances scheduler time but does not update register backgrounds;
- direct operations should receive `time=now(sim)` when scheduler time is intended;
- omitted time uses the existing selected-slot local-time synchronization behavior;
- sampled observables and measurements evolve backgrounds only when given the intended time;
- plain `traceout!` does not advance time;
- helpers without a time keyword require an explicit `uptotime!`;
- a top-level symbolic tensor can initialize as separate stored factors;
- wrapping the complete tensor in `SProjector` or `MixedState` creates one joint symbolic state;
- `apply!` stores a joint composition, while `observable` composes temporarily.

The low-level purifier advances all four participating slots to `now(sim)` before invoking its circuit helper.

## Clifford trajectories, noise, and protocols

- A general convex Clifford mixture and supported sampled Clifford background events produce one sampled trajectory; ensemble curves require multiple independent runs.
- Rank-deficient stabilizer mixtures can still be represented exactly.
- A variable named `T1` has no effect on registers configured only with `T2Dephasing(T2)`.
- `QuantumOpticsRepr` can use `T1T2Noise(T1, T2)` when both processes are intended; `CliffordRepr` does not currently support that amplitude-decay model.
- Use `@process prot()` for fire-and-forget scheduling and `@yield @process prot()` when a resumable caller must wait for one round.
- Do not request repeated rounds on fixed occupied communication slots unless another process clears those slots.

## Design choice

Selected: correct the maintained executable examples and place precise backend/time/modeling guidance next to the affected public interfaces.

Rejected: adding the full pasted seven-qubit simulation, teaching native QuantumClifford operations as the primary API, adding exported names, or invoking the credentialed documentation deployment path locally.

## Validation

- All seven `firstgenrepeater_lowlevel` example test files under `xvfb`: 9 passed.
- Full `general` suite: 14,767 passed.
- Focused snippets checked scheduler/register time, tensor/projector factorization, and one-round protocol waiting.
- An identity-containing Clifford Pauli snippet passed with the QuantumSymbolics 0.4.17 candidate; that release is now registered, and [#524](https://github.com/QuantumSavory/QuantumSavory.jl/pull/524) adds the public QuantumSavory regression.
- The credentialed/deploying `docs/make.jl` path was intentionally not run locally; documentation CI will run the supported build.

## Dependencies and merge order

- This PR makes no dependency edit itself.
- It does not require [QuantumClifford.jl #776](https://github.com/QuantumSavory/QuantumClifford.jl/pull/776) / QuantumClifford 0.11.7.
- Its complete documented symbolic route depends on merged [QuantumSymbolics.jl #218](https://github.com/QuantumSavory/QuantumSymbolics.jl/pull/218), **registered QuantumSymbolics 0.4.17**, and [QuantumSavory.jl #524](https://github.com/QuantumSavory/QuantumSavory.jl/pull/524).

Keep this PR in draft until:

1. [#524](https://github.com/QuantumSavory/QuantumSavory.jl/pull/524) merges;
2. this branch is refreshed from `master` after merged #518 through #521 and #524; and
3. its example/documentation checks are rerun on the refreshed head.

## Registration sequence

Do not register QuantumSavory from this branch. After this PR and every issue-301 behavior/dependency PR merge, a separate QuantumSavory `0.7.1` release PR must update the version, date the changelog, verify registry-only dependency resolution, and run the final smoke/test matrix. **QuantumSavory 0.7.1 is registered only after that release PR merges.**

This PR references #301 but does not close it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## v0.7.1 - unreleased
 
+- **(fix)** Correct the Bell-fidelity parameter in the low-level repeater
+  example and clarify backend-neutral operations, register time, factorization,
+  Clifford trajectories, and one-round entangler scheduling in the guides.
 - Indexed Gaussian operations now delegate to Gabs' canonical
   `apply!(state, indices, operation)` API; Gabs 1.3.8 is now required.
 - Simulation, network, protocol, and visualization traces now use
--- a/docs/src/API_ProtocolZoo.md
+++ b/docs/src/API_ProtocolZoo.md
@@ -25,6 +25,20 @@ prot = EntanglerProt(sim, net, 1, 2)
 @process prot()
 ```
 
+`@process prot()` schedules the protocol and lets the caller continue. Inside
+an `@resumable` caller, wait for a one-round protocol when later work requires
+the generated pair:
+
+```julia
+prot = EntanglerProt(sim, net, 1, 2; rounds=1)
+@yield @process prot()
+```
+
+If `chooseslotA` and `chooseslotB` select fixed communication slots, do not
+request further rounds unless another process consumes or clears the generated
+pair between rounds. Further rounds otherwise wait for those occupied slots to
+become free.
+
 This matters because the protocol object packages:
 
 - the simulation handle,
--- a/docs/src/backgrounds.md
+++ b/docs/src/backgrounds.md
@@ -8,9 +8,9 @@ is present whether or not someone is actively touching that subsystem.
 reg = Register([Qubit(), Qubit()], [T2Dephasing(10.0), nothing])
 ```
 
-This is a declarative noise model. You state what process is present, once, at
-model construction time. You do not manually weave noise updates through every
-gate, wait, and measurement in the protocol code.
+This is a declarative noise model. You state what process is present once at
+model construction time. Register operations apply it when they advance a
+slot to a requested time.
 
 ## Why This Matters
 
@@ -22,14 +22,15 @@ If noise were modeled by hand, every protocol would need custom bookkeeping for
 "advance the state, then apply the operation, then advance it again." That is
 error-prone and it makes protocol code much harder to read.
 
-QuantumSavory keeps that bookkeeping in the framework instead.
+QuantumSavory keeps the evolution bookkeeping in the register operations.
+Discrete-event waits do not update register states by themselves.
 
 ## Time Evolution Is Demand Driven
 
-Each subsystem carries its own local simulation time. When a protocol applies a
-gate, requests an observable, or otherwise touches part of the state,
-QuantumSavory advances the relevant subsystem to the requested time before
-continuing.
+Each subsystem carries its own local simulation time. Operations advance the
+relevant state only when their contract requests a later time. In a
+`ConcurrentSim` process, pass `time=now(sim)` to direct register operations
+when scheduler time is intended. A `timeout` advances only the scheduler.
 
 This means:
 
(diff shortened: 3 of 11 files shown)


Krastanov:
this is a bit sloppy and disorganized

Issue #531, opened 2026-08-13
Add `show` methods for qTCP protocols following these guidelines


LBacciottini:
This issue provides the specifics for the content that the `show` method should produce for `LinkController`, `NetworkNodeController`, `EndNodeController`, i.e. the protocols composing the qTCP suite. For the rest of this issue, I'll describe the figure canvas returned by `show` for each of these protocols, and what that canvas is composed of.

## `LinkController`

- The header of the canvas should identify the link, i.e. the two nodes.
- The canvas should contain an inset figure showing (part of) the `show` canvas from the `EntanglerProt` that is internally being used by the `LinkController`, to give insight into the entanglement generation part of the protocol.
- The rest of the canvas is dedicated to the `LinkController` as a queue server:
    - One plot shows an empirical pdf of `LinkLevelRequest` interarrival times for each one of the two nodes (so two subfigures, one next to the other).
    - One plot shows the empirical pdf of `LinkLevelRequest` sojourn times from the time they arrive to the time they are processed and deleted. This plot can be cumulative since the `LinkController` has a single queue of requests. Both `LinkController` plots should also report the sample mean and median values.

### Mock up

<img width="338" height="583" alt="Image" src="https://github.com/user-attachments/assets/af54c562-db2f-4571-94e2-d844f78d75aa" />

## `NetworkNodeController`
- The header of the canvas should identify the node this protocol is running on
- The entirety of the canvas should contain the following informative plots:
    - One plot showing a per-neighbor backlog of enqueued `QDatagram`s as the simulation proceeds, i.e. simulation time on the x axis. This plot should thus update live as the simulation runs. This should be one plot figure with k lines, one per `flow_id`.
    - One table with the average sojourn time of `QDatagram`s, indexed by `flow_id`. This table should be placed next to the above plot and complement its information by giving quantitative info about how much time, on average, a `QDatagram` spends in the queue.
    - One plot showing the total number of processed (i.e. swapped) `QDatagram`s for each `flow_id`.

### Mock up

<img width="648" height="548" alt="Image" src="https://github.com/user-attachments/assets/bce6f1a2-22dc-4b15-bc69-219ccd9f655a" />

## `EndNodeController`
- The header of the canvas should identify the node this protocol is running on
[...]

PR #534, opened 2026-08-14
Add LinkController protocol visualizations

Krastanov-agent:
## Summary

- record LinkController request arrivals and completed sojourn times in one private typed log
- add a semantic `text/html` display with the existing protocol class conventions and embedded EntanglerProt display
- add a Makie `image/png` display with the embedded entangler, per-endpoint interarrival PDFs, and aggregate sojourn PDF
- report sample counts, means, and medians, including explicit empty-sample handling
- centralize protocol PNG headers and scale figure height through a `protshowrows` trait

This implements only the LinkController subsection of [issue #531](https://github.com/QuantumSavory/QuantumSavory.jl/issues/531). It does not close the issue; the other QTCP controllers are separate work.

## Review order

1. `d5466f97` records request timing state and preserves existing constructors.
2. `a6f4a9e7` adds the HTML display and its contract tests.
3. `93a88b02` adds the PNG display and its Cairo smoke tests.
4. `ccd22509` improves the EntanglerProt distribution cutoff and uses bars for short tails.
5. `1703eec3` removes the misplaced changelog entry and relies on the shared `AbstractProtocol` PNG show method.
6. `b4024691` documents the private core and HTML helpers.
7. `b8d7afff` documents the private Makie helpers.
8. `14cbe2f0` centralizes protocol PNG labels and row-based figure sizing.
9. `827969b6` declares the four LinkController plot rows and removes its duplicate label.

## Validation

- `general/protocolzoo_qtcp`: 37/37 passed
- `general/protocol_show_html_contracts`: 37/37 passed
- `general/interactiveutils`: 44/44 passed
- default `general` shard: 14,858/14,858 passed
- `plotting/show_png`: 10/10 passed with CairoMakie
- `git diff --check`: passed
- initial three-commit implementation independently reviewed with no blockers
- populated PNG and HTML outputs inspected manually

The full plotting shard was not run. GLMakie cannot initialize on this headless host because `DISPLAY` is unavailable; the requested Cairo `image/png` path passed.

## Renderings

The following outputs use a two-node QTCP request simulation. The HTML screenshot applies a small consumer stylesheet to the semantic tags emitted by `show`.

### `image/png`

LinkController image/png rendering

### `text/html`

LinkController text/html rendering

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
   Julia's standard logging macros with stable `event` symbols and structured
   metadata. Logs expose stable, filterable domain groups through `LOG_GROUPS`.
 - Additional visualization methods for states of registers.
+- QTCP `LinkController` protocols now provide HTML and PNG summaries of
+  per-endpoint request interarrival times and aggregate request sojourn times.
 - Dense observables on pure Clifford states now warn before converting the entire
   stabilizer state to an exponentially sized ket.
 - `QuantumMCRepr` is supported much more thoroughly, providing a faster alternative to `QuantumOpticsRepr` that is just as general.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,6 @@
   Julia's standard logging macros with stable `event` symbols and structured
   metadata. Logs expose stable, filterable domain groups through `LOG_GROUPS`.
 - Additional visualization methods for states of registers.
-- QTCP `LinkController` protocols now provide HTML and PNG summaries of
-  per-endpoint request interarrival times and aggregate request sojourn times.
 - Dense observables on pure Clifford states now warn before converting the entire
   stabilizer state to an exponentially sized ket.
 - `QuantumMCRepr` is supported much more thoroughly, providing a faster alternative to `QuantumOpticsRepr` that is just as general.
--- a/ext/QuantumSavoryMakie/show_protocol.jl
+++ b/ext/QuantumSavoryMakie/show_protocol.jl
@@ -83,6 +83,8 @@ function _link_timing_histogram(subfig, samples; title)
     return axis
 end
 
+protshowrows(::QuantumSavory.ProtocolZoo.QTCP.LinkController) = 4
+
 function protshowimage(subfig, prot::QuantumSavory.ProtocolZoo.QTCP.LinkController)
     qtcp = QuantumSavory.ProtocolZoo.QTCP
     label_a = compactstr(prot.net[prot.nodeA])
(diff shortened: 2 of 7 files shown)


Krastanov-agent:
We will be adding new visualizations and `show` methods to the QTCP protocol components in QuantumSavory.ProtocolZoo.

See docs.makie.org for the plotting library we will be using.

See ext/QuantumSavoryMakie/show_protocol.jl for how the image/png show method is implemented for a few existing protocols (specifically `protshowimage` which is then called from the actual `show`). Use default styling as much as possible for your implementations.

See src/ProtocolZoo/show.jl for how the text/html show methods are implemented. Importantly, keep the pattern of tag classes consistent so that they can be semantically styled by a consumer of the `show` method. You probably will need to create src/ProtocolZoo/qtcp/show.jl to simplify module namespacing.

Your task is to implement png and html show methods for just the LinkController of the QTCP protocols (other agents will implement the rest). Read this issue for suggested content and a mockup https://github.com/QuantumSavory/QuantumSavory.jl/issues/531

You might or might not need to introduce private `._log` field or similar private fields to the LinkController struct. If such a structure is necessary keep it as simple as possible. Investigate other protocols with private fields as an example.

Have changes to the protocols and each show method in separate commits to ease review. Simplicity of your code and simplicity of commit-by-commit review is paramount.  When done, post a PR and on that PR post this exact prompt so that the reviewer knows how this work started. In the PR post a rendering of the image/png you implemented and a rendering of the text/html as well, e.g. by using the existing QTCP example.

Krastanov-agent:
Implemented in [`ccd22509`](https://github.com/QuantumSavory/QuantumSavory.jl/commit/ccd22509490e8b8de817ca04f6fd53812c03fc23).

## Updated `image/png` rendering

Updated LinkController image/png rendering

## Follow-up prompt

> Modify how the EntanglerProt renders its PNG. Cut off the x axis at n+2 where for n the probability is under 0.001. If n is less than 10, use bar plots instead of steps. Have all of this as a clear simple well-documented easy to review new commit. Render the example PNG in the current PR description again when done and post it in a new comment that also copies this prompt itself, as a self-documentation.

Krastanov-agent:
Implemented in two focused commits:

- [`b4024691`](https://github.com/QuantumSavory/QuantumSavory.jl/pull/534/commits/b4024691) documents the private core and HTML helpers.
- [`b8d7afff`](https://github.com/QuantumSavory/QuantumSavory.jl/pull/534/commits/b8d7afff) documents the private Makie helpers.

Updated `image/png` rendering from the synchronized PR head:

Updated LinkController image/png rendering

Prompt:

```text
I made some changes to the remote PR branch (removed the changelog as it was in the wrong section, and removed the show method because the method for QuantumSavory.ProtocolZoo.AbstractProtocol should cover it). Make sure you sync up with it. Then do the following final changes (and follow the same pattern above of posting the prompt and the updated PNG renderings, after making new well documented easy to review commits):

  - for each of the private helper functions you have introduced, add an extremely short single line docstring explaining their role
```

Krastanov-agent:
Implemented in two focused commits:

- [`14cbe2f0`](https://github.com/QuantumSavory/QuantumSavory.jl/pull/534/commits/14cbe2f0) adds the documented `protshowrows` fallback, shared introspected header, and row-scaled figure construction.
- [`827969b6`](https://github.com/QuantumSavory/QuantumSavory.jl/pull/534/commits/827969b6) assigns four rows to LinkController and removes its duplicate plot label.

The populated example now renders through the generic `AbstractProtocol` `image/png` method at 1200×3700 pixels:

Updated LinkController row-scaled image/png rendering

<details>
<summary>Prompt</summary>

```text
It seems we need to make a few more changes, following the very same pattern of updating the visualization and posting the prompt being used. Introduce a new function protshowrows that will be used by the show method to decide how many rows of plots to put in the figure and scale the height accordingly. Have it default to returning `1`. The width will always be the same, whatever the current default is. Always leave a little extra at the top in addition to these rows for the label. Move the label plotting to the show method itself using various introspection features of the protocol so that we do not need to re-implement the label in every single plot.
```

</details>

PR #587, opened 2026-09-17
Register MessageBuffer waiters before onchange returns (fix a lost wake-up)

giovanniligato:
Fix a lost wake-up in `onchange(::MessageBuffer)`.

**The bug.** `onchange(mb)` returned `@process wait_process(...)`: the caller was registered as a waiter (or consumed a queued `no_wait` token) only when that process ran, one scheduler step after the caller had yielded. If a message was delivered in that step by a local `put!(mb, …)` while another task was already blocked on the same buffer, the delivery saw `nbwaiters == 1`, fired the notifier for the already-blocked task, and stored no token. The already-blocked task woke, found nothing for itself, and waited again; the caller then registered on the fresh generation and blocked on a change that had already happened. Its message sat in the buffer and nothing woke it unless an unrelated message arrived later. `onchange(::Register)` does not have this window because `lock(::ChangeNotifier)` counts the waiter synchronously.

**The fix.** `onchange(::MessageBuffer)` (and the deprecated `wait(::MessageBuffer)`) now registers the caller before returning: it either consumes one queued token and returns a process that finishes as soon as it is scheduled, or attaches to the current notifier generation through `lock(mb.tag_waiter)`. The return type stays `ConcurrentSim.Process` in both branches, and the queued-token behaviour for arrivals with no waiter is unchanged. The blocking path is one scheduler step shorter, since the intermediate `wait_process` is gone.

**Reproducer** (`test/general/messagebuffer_tests.jl`, "an arrival right after onchange is not lost while another task is blocked"): a bystander is blocked on the buffer; at t = 1 the victim queries, finds nothing and yields on `onchange(mb)`; later in the same step the victim's message is delivered with `put!(mb, …)`; the bystander's own message arrives at t = 5. On `master` the victim wakes at t = 5 (`("victim", 1.0) in wake_log` evaluates to `("victim", 1.0) in [("victim", 5.0), ("bystander", 5.0)]`); with this change it wakes at t = 1. The ordering is deterministic under ConcurrentSim's (time, priority, insertion id) event order.

**Validation.** On Julia 1.12.5, Windows: the new test fails on `master` and passes with the fix; `Pkg.test(; test_args=["general", "--jobs=2"])` passes all 15,251 assertions across 65 files (Aqua included, 8 min 37 s); the `jet` shard passes; `git diff --check` is clean. The plotting, examples and docs shards were not run, since the change does not touch them.

**CHANGELOG** (also added under "Unreleased" in the file):

- **(fix)** `onchange(::MessageBuffer)` now registers the caller as a waiter before returning. Previously the registration happened inside a separately scheduled process, so a message delivered later in the same simulation step, after the caller had yielded but before that process ran, while another task was already blocked on the same buffer, woke only that other task and was never noticed by the caller (a lost wake-up that could block a protocol forever).

**AI usage disclosure** (per the [AI usage and disclosure policy](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md); the points below are mine, I only asked the model to polish the English)

1. **Agent, model, version.** Claude Code with Claude Fable 5.1.
2. **How the model was used.** Here is the workflow I followed. I have been developing a distributed entanglement-distribution protocol on top of QuantumSavory. The protocol runs several processes on the same network node, and they all wait on that node's message buffer with the usual `querydelete!` + `@yield onchange(mb)` loop. Occasionally one of them never woke up, even though its message was sitting in the buffer, and the whole run stalled; the pattern was exactly the one described under "The bug" above: the process that hung had just started waiting while another process on the same node was already blocked on the buffer. At first I only worked around it in my protocol. I then asked the model to formalize the failure precisely and to determine whether the fault was in my protocol or in the simulator, and it traced it to the deferred registration of the waiter in `onchange(::MessageBuffer)`. After that I asked it to propose a solution, which it did together with the regression test; I studied the proposal, understood it, and confirmed it.
3. **What the model produced incorrectly and what I changed by hand.** Nothing: the proposed change solved the problem and I did not have to modify it.
4. **Reading the code.** I have read and understood every modification in the commit.
5. **Running the tests.** I ran the tests locally myself, and they all passed.

---


Krastanov:
Thanks, @giovanniligato ! I will try to look through this later in the week!

